PR #11, opened 2024-09-12
Delete all liquifier work and move to new repo

drewstone:
CC @shekohex this should make building gadget faster, so it's not pulling in so many modules.


shekohex (approved):

drewstone:
@shekohex can you make sure to update the template/gadget repos so we're pulling in less deps here?

PR #18, opened 2024-11-11
Cross chain restaking system

drewstone:
# Overview
This adds a cross chain restaking system for Tangle.

**Asset depositing and delegation flow**
1. Users deposit any ERC20 and lock it on a remote chain. The user can select which bridge that want to use to relay the message that this has occurred.
2. Tangle receives this message on one of the `Adapters`, such as `HyperlaneAdapter` or `RouterAdapter` or another bridge adapter eventually. This event contains `ICrossChainDepositMessage.AssetMessage`.
3. The adapter routes this message to the `CrossChainAssetVault` which processes the `AssetMessage`. The protocol for handling is to mint a `SyntheticRestakeAsset` which contains metadata about the asset locked on the remote chain. It acts like a soulbound ERC20 (it doesn't ever leave the `CrossChainAssetVault`, it simply changes owners).
4. The `CrossChainAssetVault` implements the `AssetDelegationBase` which allows the vault to interact with the `MultiAssetDelegation` precompile, the foundational precompile for interacting with the runtime's restaking infrastructure.

**Slashing and Blueprint development flow**
The reason for cross-chain functionality is allowing any asset to restake on operators and in turn blueprints. The cross-chain information relevant for Blueprints is many fold. We need to deal with **slashing**, **rewards**, **job results**, and maybe even more.
1. The `XCBlueprintServiceManager` is a cross-chain enabled Blueprint. The idea is to relay the calls for slashing, rewards, and job results to remote chains. This contract uses the `CrossChainBridgeManager` to dispatch messages to remote chains. The bridge stores bridges which each have an implemented messsenger using `ICrossChainMessenger` interface.
2. On remote chains, we deploy `BlueprintReceiver` contracts, notably `HyperlaneBlueprintReceiver` and `RouterBlueprintReceiver` contracts which receive the messages containing information about slashes, rewards, and job results. Each contract implements `BaseBlueprintReceiver` which handles cross-chain messages in a standard format so we don't have to care which bridge is used. The intention is that a user who uses Hyperlane to bridge their data should receive notifications of slashes/rewards over Hyperlane too.
3. On the remote chain, we receive these messages and we handle them accordingly. What's left is to have the `BlueprintReceiver` contracts actually trigger the slash handler on the `RemoteAssetVault` (this hasn't been created yet).

**Interacting with the precompile on Tangle**
- `AssetDelegationBase` is the contract we will want to standardize as a way for contracts with assets to interact with the delegation system. This should be our way for LRTs to integrate with Tangle, simply implement this contract system. CC @1xstj 

# Diagram
image


drewstone:
Todos:
- replace messengers and adapters with Glacis to manage xc messaging. 
- add a testing scaffold for testing this even if it fails or stubs out most of the interactions.

PR #44, opened 2024-12-09
feat: rust bytecode crate

Tjemmmic:
# Overview

Adds a Rust crate that exports the bytecode of specified contracts for simplified use of tnt-core in Rust.


drewstone:
LG

drewstone (approved):

PR #50, opened 2025-01-22
Update MSBM w/ new features in Tangle (same branch name)

drewstone:


shekohex:
@drewstone you will need to add more testing here to increase the coverage so we can merge this and publish a new version of tnt-core.

PR #57, opened 2025-04-22
feat: rpc address update event

Tjemmmic:
# Overview

Adds event for updating an Operator's rpc address in their preferences. Also removed PriceTargets.


Serial-ATA (approved):

Serial-ATA:
You should also bump the bytecode crate version

PR #58, opened 2025-05-15
feat: quality of service

Tjemmmic:
# Overview

Implements support for Heartbeats that ensure operators are running their service instances.


drewstone (commented):
A blueprint can override these default values correct?

drewstone (approved):

PR #83, opened 2026-02-12
feat: cloud earnings QA - payment distribution events and local env improvements

vutuanlinh2k2:
## Summary
- **Payment distribution events**: Add `PaymentDistributed` and `OperatorRewardDistributed` events to `Payments.sol` and corresponding indexer handlers to track payment splits (developer, protocol, operator, staker) for cloud earnings QA
- **Local env improvements**: Extend `start-local-dev.sh` and `LocalTestnet.s.sol` with multi-instance support, subscription mode, service-leavable options, and a new `submit-job-result.sh` script for end-to-end job lifecycle testing

## Test plan
- [ ] Verify payment distribution events are emitted correctly during job result submission
- [ ] Verify indexer correctly captures and indexes `PaymentDistributed` and `OperatorRewardDistributed` events
- [ ] Test local env scripts with multi-instance and subscription mode options
- [ ] Run `submit-job-result.sh` for full job lifecycle (submit → result → payment distribution)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/indexer/config.local.yaml
+++ b/indexer/config.local.yaml
@@ -38,7 +38,9 @@ networks:
           - event: Initialized(uint64 version)
           - event: JobCompleted(uint64 indexed serviceId, uint64 indexed callId)
           - event: JobResultSubmitted(uint64 indexed serviceId, uint64 indexed callId, address indexed operator, bytes output)
+          - event: JobEventRateSet(uint64 indexed blueprintId, uint8 indexed jobIndex, uint256 rate)
           - event: JobSubmitted(uint64 indexed serviceId, uint64 indexed callId, uint8 jobIndex, address caller, bytes inputs)
+          - event: JobSubmittedFromQuote(uint64 indexed serviceId, uint64 indexed callId, uint8 jobIndex, address caller, address[] quotedOperators, uint256 totalPrice, bytes inputs)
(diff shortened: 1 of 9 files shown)


vutuanlinh2k2:
I re-checked Claude’s feedback against the current implementation. I think there is **one concrete issue worth fixing before merge**:

- `script/sh/local-env/submit-job-result.sh` callId extraction can pick the wrong log.
  - Current logic scans for the first log with `>=3` topics and treats `topics[2]` as `callId`.
  - In `submitJob`, payment-related events can be emitted before `JobSubmitted`, so `topics[2]` from an earlier log may be an address/value, not the job call ID.
  - This can produce a wrong callId in script output and break follow-up `result` submissions.
  - Suggested fix: filter logs by `topic0 == keccak256("JobSubmitted(uint64,uint64,uint8,address,bytes)")`, then decode `topics[2]`.

For the other points:
- `PaymentDistributed.operatorPoolAmount/restakerPoolAmount` are consistent with the current pool split logic (not seeing an accounting mismatch).
- Indexer service/operator existence checks look intentionally relaxed because these relations are nullable in schema; skipping events on missing parent may lose useful history.
- Script complexity/documentation notes are valid maintainability feedback but not blockers.

drewstone (approved):

PR #103, opened 2026-04-20
chore(base-sepolia): prep fresh testnet launch

drewstone:
## Summary

- Preps a fresh Base Sepolia TNT protocol launch owned by the new testnet deployer wallet.
- Enables Base Sepolia USDC in the launch config.
- Adds a dedicated Base Sepolia deploy wrapper that defaults to the `deploy_size` Foundry profile so `ServiceFeeDistributor` stays under the EIP-170 contract size limit.
- Restores backward compatibility in the generated manifest by emitting both `staking` and `restaking` with the same `MultiAssetDelegation` address.

## Change Class

- `Class A` (docs/tooling only)
- `Class B` (single-crate behavior)
- `Class C` (cross-crate/runtime behavior)
- `Class D` (protocol/security/metadata semantics)
- Selected class: Class C
- Why this class: this changes protocol deployment configuration, generated manifest compatibility, and the operational launch path for Base Sepolia.

## Behavior Contract

- Current behavior: Base Sepolia launch config still pointed at an older admin wallet, left USDC disabled, and required operators to remember an unstated `deploy_size` requirement to avoid the `ServiceFeeDistributor` contract-size failure. Generated manifests also exposed only `staking`, while some downstream automation still expected `restaking`.
- Intended behavior: a fresh Base Sepolia deployment should be launchable by the current deployer wallet, include the intended USDC asset, default to a safe deployment profile, and produce manifests that work with both newer and older consumers.
- Invariants that must hold: `deploy/config/base-sepolia.json` must bootstrap to `0x03A763af11dC58B459A469A9895389F5880B45F4`; Base Sepolia USDC must resolve to Circle's official testnet contract; `script/sh/deploy-testnet-base-sepolia.sh` must run with `FOUNDRY_PROFILE=deploy_size` by default; generated manifests must include `tangle`, `staking`, `restaking`, and `statusRegistry` consistently.
- Failure mode choice (`fail-closed` or `fail-open`) and rationale: fail-closed. Launch scripts should refuse missing RPC/private-key configuration rather than silently using a broken deployment path.

## Risk and Scope

- Security impact: medium-low. This does not change protocol runtime logic, but it changes who owns a fresh testnet deployment and how deployment artifacts are consumed.
- Compatibility impact (APIs, metadata format, configs): low-positive. The manifest adds a backward-compatible `restaking` alias without removing `staking`.
- Migration notes (if any): use the new wrapper script or explicitly set `FOUNDRY_PROFILE=deploy_size` when deploying to Base Sepolia.
- Rollback plan: revert the config updates, manifest alias, and wrapper script.

## Verification

```bash
cast balance 0x03A763af11dC58B459A469A9895389F5880B45F4 --rpc-url https://sepolia.base.org
# 0.0037 ETH

cast call 0x036CbD53842c5426634e7929541eC2318f3dCF7e "balanceOf(address)(uint256)" 0x03A763af11dC58B459A469A9895389F5880B45F4 --rpc-url https://sepolia.base.org
# 10 USDC

cast block-number --rpc-url wss://base-sepolia-rpc.publicnode.com
cast chain-id --rpc-url wss://base-sepolia-rpc.publicnode.com
# succeeded; public WebSocket endpoint is usable

forge inspect src/rewards/ServiceFeeDistributor.sol:ServiceFeeDistributor deployedBytecode | ...
# default profile => 24601 bytes (too large)

FOUNDRY_PROFILE=deploy_size forge inspect src/rewards/ServiceFeeDistributor.sol:ServiceFeeDistributor deployedBytecode | ...
# deploy_size => 24251 bytes (under 24576 limit)

FOUNDRY_PROFILE=deploy_size forge script script/FullDeploy.s.sol:FullDeploy --rpc-url https://sepolia.base.org --non-interactive -vv
# launch simulation in progress / expected to validate the deploy_size path

bash -n script/sh/deploy-testnet-base-sepolia.sh
# passed
```

## Harness Evidence

- Reproducer added/updated: no new test harness; validation is operational via live Base Sepolia simulation.
- Negative-path coverage added: wrapper script enforces missing envs and safe default profile.
- Key assertions that prove the fix: launch config targets the active deployer wallet, USDC is enabled, `deploy_size` removes the EIP-170 blocker, and the manifest regains `restaking` compatibility.

## Checklist

- [x] Reviewed the change against the PR template requirements.
- [x] Validated shell script syntax.
- [x] Validated live Base Sepolia wallet funding and token state.
- [x] Validated a working public Base Sepolia WebSocket endpoint.
- [x] Validated `ServiceFeeDistributor` bytecode size under `deploy_size`.


drewstone:
Launch facts validated during prep:

- Base Sepolia deployer `0x03A763af11dC58B459A469A9895389F5880B45F4` currently has `0.0037 ETH`, `10 USDC`, and nonce `0`.
- Base Sepolia public WebSocket endpoint validated from CLI: `wss://base-sepolia-rpc.publicnode.com`.
- The old naming mismatch is real: downstream automation may still refer to `RESTAKING_CONTRACT`, but that maps to the `MultiAssetDelegation` address. This PR makes the generated manifest emit both `staking` and `restaking` for compatibility.
- FullDeploy simulation on Base Sepolia succeeded under the default profile, but `ServiceFeeDistributor` is too large there (`24601` bytes). `deploy_size` brings it down to `24251`, which is why the new wrapper defaults to `FOUNDRY_PROFILE=deploy_size`.

This means testnet ETH is not the blocker anymore. The launch-critical path is now the `deploy_size` deployment path and then wiring the resulting `tangle` / `restaking` / `statusRegistry` addresses into the blueprint operator flow.

PR #115, opened 2026-05-04
fix(security): pre-mainnet hardening across deploy, MBSM, beacon, oracles, quotes, BLS

drewstone:
## Summary

Closes 12 critical and several high-severity findings ahead of Base mainnet launch. All 1472 existing tests still pass; 7 new regression tests cover the ship-blockers under `test/security/`.

This is the code half of the launch readiness work. Operational follow-ups (populating `base-mainnet.json`, deploying/pinning Timelock+Governor, ISM/DVN pinning) are listed at the bottom.

### Deploy
- `FullDeploy._requireProductionRoles` refuses to broadcast on mainnet chain ids when `admin/treasury/timelock/multisig` are zero, equal the deployer, or `revokeBootstrap=false`. Bypassable on local/test via `TANGLE_DEPLOY_LOCAL=1`.

### MBSM
- Synchronous `deprecateVersion` deleted; replaced by `queueEmergencyDeprecation` + 24h delay + `executeEmergencyDeprecation` + `cancelEmergencyDeprecation`. A single MANAGER_ROLE call can no longer brick every pinned blueprint.
- `addVersion` now rejects EOAs (`code.length > 0`).

### Beacon
- `ValidatorPod`: `0x02` (Pectra compounding) credentials raise the per-validator cap to 2048 ETH (`MAX_EFFECTIVE_BALANCE_GWEI_02`); `0x01` keeps 32 ETH.
- `BeaconChainProofs`: state tree height `5->6`, validator/balance gindex `43/44->75/76` for post-Pectra layouts.
- `L2SlashingReceiver`: duplicate `(sourceChainId, sender, nonce)` reverts with `NonceAlreadyProcessed` instead of silently returning so the relayer can detect replay.
- `L2SlashingConnector`: per-destination `lastProcessedSlashingFactorByChain`, `cumulativeSlashAmountByChain`, and `nonceByChain`. Multi-destination relays now actually work.

### Staking / vault
- `LiquidDelegationVault.redeem` takes an explicit `requestId` (ERC-7540 conformant). Share-equality matching removed.
- `StakingDelegationsFacet.executeDelegatorUnstakeAndWithdraw` reverts with `PendingSlashExists` when the operator has any pending slash, closing the vault path that bypassed the M-9 invariant.
- `AssetAdapterFactory` switched to `Ownable2Step`.

### Oracles
[...]

PR #116, opened 2026-05-04
fix(security): timelock storage slot, drip mutex, inflation pool deregister

drewstone:
## Summary

Follows up #115 with the code-level remaining items from the pre-mainnet audit. Three operational items (populate `base-mainnet.json` multisigs, deploy/pin Timelock+Governor, choose Hyperlane ISM / LayerZero DVN config) need your input and aren't in this PR.

### TangleTimelock — fix `_setMinDelay` slot + restore `onlySelf`
The previous override of `updateDelay` had two bugs:
1. **Authorization regression** — relaxed parent's `onlySelf` to `onlyRole(DEFAULT_ADMIN_ROLE)`. Compromised admin could change the delay outside a queued proposal.
2. **Wrong storage slot** — `sstore(0x33, newDelay)`. OZ 5.x `TimelockControllerUpgradeable` uses ERC-7201 namespaced storage at `0x9a37c2aa…fb3600`; `_minDelay` is at that location + 1. Writing to slot 51 was a silent no-op for `getMinDelay()` and could corrupt unrelated proxy storage.

Now:
- `onlySelf` reinstated (callable only via a queued, scheduled proposal).
- `_setMinDelay` writes to `TIMELOCK_CONTROLLER_STORAGE_LOCATION + 1`.
- Bounds [1 day, 30 days] still enforced.

### StreamingPaymentManager — drip mutex
`dripAndGetChunk` and `dripOperatorStreams` are now `nonReentrant`. Closes the same-tx double-drip race where the transfer-to-distributor hook could re-enter and process additional chunks before the first frame finished updating `lastDripTime`.

### InflationPool — deregister inactive operators
`trackedOperators` previously grew monotonically; every distribution iterated stale entries. New paths:
- `deregisterOperator(address)`
- `deregisterOperators(address[])`
- `deregisterCustomer(address)`
- `deregisterDeveloper(address)`

Each clears the `isTracked*` flag, swap-and-pops the address out of the tracked list, and clears `*RegistrationEpoch`. Pending rewards remain claimable. Re-registration is permitted.

## Tests

Two new files under `test/security/`:
- `TimelockSetMinDelayTest.t.sol` (3 tests): non-self caller reverts, `getMinDelay()` reflects new value, bounds enforced.
- `InflationPoolDeregisterTest.t.sol` (3 tests): removes from tracked list, non-admin reverts, re-registration allowed.

**Full suite: 1478/1478 passing** (1472 baseline + 6 new).

## Test plan

- [x] `forge build` clean
[...]

PR #117, opened 2026-05-04
feat(contracts): TEE-attested service approval commitments

drewstone:
## Summary

Workstream **B + C** from the [Tangle Cloud app-store program plan](https://github.com/tangle-network/dapp/blob/drew/tangle-cloud-cross-page-redesign/docs/blueprint-app-store-program-plan.md):

- **B — `AppRegistry.sol`** (greenfield) — on-chain home for blueprint apps: publisher namespaces, verified flag, per-app metadata pointers with strict version monotonicity, full version history, two-step namespace transfer, governance-only verify/revoke. UUPS-upgradeable + AccessControl.
- **C — TEE-attested `OperatorSecurityCommitment`** — operators commit to a specific TEE backend + measurement + nonce-binding at approval time; the live attestation produced when the workload provisions can be cross-checked against the on-chain commitment. `DirectTdx` is rejected at approval (mirror of `ai-agent-sandbox-blueprint` PR #50 policy).

Builds on the security primitives PR #115 merged earlier today (`MBSMRegistry`, BLS commitment paths in `ServicesApprovals`, `OperatorSecurityCommitment` types).

### Storage

- `mapping(bytes32 => address) _namespaceOwner` — namespace → owning publisher.
- `mapping(bytes32 => bool) _namespaceVerified` — governance-set verification flag.
- `mapping(bytes32 => NamespaceTransfer) _pendingTransfers` — two-step transfer state.
- `mapping(bytes32 => AppRecord) _apps` — per-app record.
- `mapping(bytes32 => AppMetadataPointer[]) _appHistory` — every version ever pinned, in order.
- `mapping(bytes32 => EnumerableSet.Bytes32Set) _namespaceApps` — O(1) listing per namespace.
- `uint256[44] private __gap` — reserved for facet-style upgrades.

### Surface

- `claimNamespace(bytes32)` — first-claimer wins.
- `transferNamespace(bytes32, address)` / `acceptNamespace(bytes32)` — two-step with stale-proposal rejection (proposer must still own the namespace at accept time).
- `publishApp(bytes32 namespace, bytes32 appId, uint64 blueprintId, AppMetadataPointer)` — publisher-only; appId must be globally unique; version != 0; metadataHash != 0.
- `pinAppVersion(bytes32 appId, AppMetadataPointer)` — strictly monotonic version; previous pointer remains in history; refreshes verified mirror; updates frozenAt.
- Governance: `setVerified(bytes32, bool)`, `revokeApp(bytes32, string reason)`.
- Views: `getApp`, `getAppHistory`, `getAppFrozenAt`, `listApps(namespace)`, `appCount(namespace)`, `namespaceOwner`, `pendingNamespaceTransfer`, `isNamespaceVerified`, `appExists`.

### Roles

- `DEFAULT_ADMIN_ROLE` administers the role hierarchy (timelock recommended).
- `UPGRADER_ROLE` authorises UUPS upgrades.
- `GOVERNANCE_ROLE` flips `verified` and revokes apps.

### Errors

`Unauthorized`, `NamespaceTaken`, `AppExists`, `AppNotFound`, `MetadataHashMismatch`, `Revoked`, `InvalidNamespace`, `InvalidAppId`, `InvalidVersion`, `NoPendingTransfer`, `StaleTransferProposal`.

### Tests (`test/registry/AppRegistryTest.t.sol`, 23 cases)

- Publish + pin + history happy path.
- Unauthorized publish (non-owner).
- Two-step namespace transfer: propose, re-propose-overwrites-prior, accept; rejection of bob's accept after carol takes over the proposal.
- Chained multi-hop transfers (alice → bob → carol) — proposer stays in sync.
- `setVerified` governance-only; `verified` flips through `getApp`.
- `revokeApp` blocks further pins; double-revoke reverts; non-governance revoke reverts.
- `pinAppVersion` rejects equal/lower versions; rejects zero hash.
- `listApps` + `appCount` via EnumerableSet (insertion order, swap-pop on remove).

### Types (`src/libraries/Types.sol`)

```solidity
enum TeeBackend {
    Phala,
    AwsNitro,
    GcpConfidential,
    AzureSkr,
    DirectTdx        // recognised but rejected at approval
}

struct TeeAttestationCommitment {
    TeeBackend backend;
    bytes32 expectedMeasurement;  // MRTD/MRENCLAVE/PCR-equivalent
    bytes32 nonceBinding;         // e.g. service request hash
    uint64 expiresAt;             // 0 = no expiry; non-zero must be future
}
```

Enum values are **append-only** by policy (storage hazard otherwise).

### Approval entrypoint

```solidity
function approveServiceWithTeeCommitments(
    uint64 requestId,
    Types.AssetSecurityCommitment[] calldata commitments,
    uint256[4] calldata blsPubkey,
    uint256[2] calldata popSignature,
    Types.TeeAttestationCommitment[] calldata teeCommitments
) external whenNotPaused nonReentrant;
```

Validates each TEE commitment (rejects `DirectTdx`, rejects past expiries), records the per-operator array under `_requestTeeCommitments[requestId][operator]` and emits `TeeCommitmentRecorded`. Activation copies the array onto `_serviceTeeCommitments[serviceId][operator]` via `TangleServicesFacet._persistTeeCommitments`.

### Read surface

```solidity
function getTeeCommitment(uint64 serviceId, address operator)
    external view returns (Types.TeeAttestationCommitment[] memory commitments);
```

Returned array matches storage order. Empty if the operator approved without TEE commitments.

### Storage layout impact

`TangleStorage.__gap` reduced **44 → 42** to accommodate the two new mappings:

```solidity
mapping(uint64 => mapping(address => Types.TeeAttestationCommitment[]))
    internal _requestTeeCommitments;
mapping(uint64 => mapping(address => Types.TeeAttestationCommitment[]))
    internal _serviceTeeCommitments;
```

### Errors

`DirectTdxNotPermitted`, `TeeCommitmentExpired(uint64 expiresAt, uint64 nowTimestamp)`, `MeasurementMismatch(bytes32 expected, bytes32 observed)`, `TeeCommitmentLengthMismatch(uint256 commitmentCount, uint256 teeCommitmentCount)`.

### Tests (`test/tangle/TeeCommitmentApprovalTest.t.sol`, 9 cases)

- Happy path: approve with single Nitro commitment, fetch via `getTeeCommitment`.
- DirectTdx rejected at approval (`DirectTdxNotPermitted`).
- Past-expiry commitment rejected (`TeeCommitmentExpired`).
- Multiple operators with distinct backends (Nitro / Phala / GCP) in one quote.
- Commitment persistence onto the activated service via `_persistTeeCommitments`.
- Empty TEE commitments array allowed (back-compat with non-TEE flows).

## Test plan

- [x] Local Solidity formatted with `forge fmt` on changed files only.
- [ ] Local `forge build` (in flight at push time on a 341-file recompile after the formatting pass; CI runs the canonical build + tests).
- [ ] CI to run `forge test --match-path 'test/registry/*'` and `test/tangle/TeeCommitment*`.
- [ ] CI to run the full existing tnt-core test suite for regressions.
- [ ] Reviewer to confirm storage gap accounting is correct (44 → 42 with two new mappings).
- [ ] Reviewer to spot-check enum append-only convention (`TeeBackend`).

## Out of scope (deferred to follow-ups, captured in the program plan)

- DCAP / Nitro / Phala verifier libraries on-chain — current design persists the commitment and assumes the off-chain provisioning oracle (or `_handleProvisionResult` hook in `ai-agent-sandbox-blueprint`) does the verifier walk. Bringing those on-chain costs significant gas; tracked under Workstream C P2 in the plan.
- `setBackendRootCa(backend, ca)` governance call for verifier root rotation — needs the on-chain verifier first.
- AppRegistry indexer schema (Envio/Hasura) for off-chain catalog listing — needs to ship coupled to this contract; tracked under Workstream B P2.
- Real AWS Nitro production validation — Workstream E in the program plan.

## Notes

This pairs with [`ai-agent-sandbox-blueprint#55`](https://github.com/tangle-network/ai-agent-sandbox-blueprint/pull/55) (Workstream E: Firecracker port mappings + auth-path GC). Together they round out the contract / runtime story for the app-store program; surface rendering + brand adoption is in [`dapp#3164`](https://github.com/tangle-network/dapp/pull/3164).

### DIFF
--- a/src/TangleStorage.sol
+++ b/src/TangleStorage.sol
@@ -370,10 +370,23 @@ abstract contract TangleStorage {
     mapping(uint64 => Types.ResourceCommitment[]) internal _requestResourceRequirements;
 
     // ═══════════════════════════════════════════════════════════════════════════
+    // TEE ATTESTATION COMMITMENTS
+    // ═══════════════════════════════════════════════════════════════════════════
+
+    /// @notice Request ID => Operator => TEE attestation commitments captured at approval.
+    /// @dev Transferred to `_serviceTeeCommitments` on activation; not used afterwards.
+    mapping(uint64 => mapping(address => Types.TeeAttestationCommitment[])) internal _requestTeeCommitments;
+
+    /// @notice Service ID => Operator => TEE attestation commitments persisted at activation.
+    /// @dev Read by blueprint contracts during their provisioning hooks to cross-check the
+    ///      live TEE attestation against the operator's on-chain commitment.
+    mapping(uint64 => mapping(address => Types.TeeAttestationCommitment[])) internal _serviceTeeCommitments;
+
+    // ═══════════════════════════════════════════════════════════════════════════
     // RESERVED STORAGE GAP
     // ═══════════════════════════════════════════════════════════════════════════
 
     /// @dev Reserved storage slots for future upgrades
     /// @dev Standard gap size is 50 slots. When adding new storage, decrease this gap accordingly.
-    uint256[44] private __gap;
+    uint256[42] private __gap;
 }
--- a/src/core/ServicesApprovals.sol
+++ b/src/core/ServicesApprovals.sol
@@ -22,6 +22,15 @@ abstract contract ServicesApprovals is Base {
     event ServiceApproved(uint64 indexed requestId, address indexed operator);
     event ServiceRejected(uint64 indexed requestId, address indexed operator);
 
+    /// @notice Emitted when an operator's TEE attestation commitment is recorded at approval.
+    /// @param requestId The service request ID being approved.
+    /// @param operator The approving operator (msg.sender).
+    /// @param backend Which TEE backend the operator commits to.
+    /// @param expectedMeasurement Measurement the live attestation must match off-chain.
+    event TeeCommitmentRecorded(
+        uint64 indexed requestId, address indexed operator, Types.TeeBackend backend, bytes32 expectedMeasurement
+    );
+
     // ═══════════════════════════════════════════════════════════════════════════
     // SERVICE APPROVAL
     // ═══════════════════════════════════════════════════════════════════════════
@@ -169,6 +178,159 @@ abstract contract ServicesApprovals is Base {
         _approveServiceWithCommitmentsInternal(requestId, commitments, blsPubkey);
     }
 
+    /// @notice Approve a service request with both security commitments, BLS public key,
+    ///         and TEE attestation commitments. Each TEE commitment is stored per-operator
+    ///         so blueprints can cross-check it against the live attestation produced when
+    ///         the workload provisions.
+    /// @dev `teeCommitments` may be empty if the request does not require a TEE workload;
+    ///      otherwise every entry MUST set a non-`DirectTdx` backend and (if `expiresAt != 0`)
+    ///      a future expiry. The list is interpreted as the operator's set of acceptable
+    ///      attestation profiles — any one matching at provisioning time satisfies the policy.
+    /// @param requestId The service request ID
+    /// @param commitments Per-asset security commitments (matches `approveServiceWithCommitments`)
+    /// @param blsPubkey BLS G2 pubkey [x0, x1, y0, y1] (zero pubkey allowed if BLS not used)
+    /// @param popSignature G1 proof-of-possession (only validated when blsPubkey is non-zero)
+    /// @param teeCommitments TEE attestation commitments to record for `msg.sender`
+    function approveServiceWithTeeCommitments(
+        uint64 requestId,
+        Types.AssetSecurityCommitment[] calldata commitments,
+        uint256[4] calldata blsPubkey,
+        uint256[2] calldata popSignature,
+        Types.TeeAttestationCommitment[] calldata teeCommitments
+    )
+        external
+        whenNotPaused
+        nonReentrant
+    {
+        // Authorize FIRST so an unauthorized caller never reaches the per-commitment
+        // SSTORE loop. Pure validation (no state change) of the commitment shape and
+        // BLS PoP follows; storage writes happen only after both gates pass.
+        _requireApprovingOperator(requestId);
+        _validateTeeCommitments(requestId, teeCommitments);
+        if (_isNonZeroBlsPubkey(blsPubkey)) {
+            _requireBlsProofOfPossession(msg.sender, blsPubkey, popSignature);
+        }
+        // Store TEE commitments BEFORE the internal approval flow triggers activation.
+        // `_approveServiceWithCommitmentsInternal` will call `_activateService` when this
+        // is the final approval, and the activation hook copies `_requestTeeCommitments`
+        // into `_serviceTeeCommitments`. Order matters: write to the request first.
+        _storeRequestTeeCommitments(requestId, msg.sender, teeCommitments);
+        _approveServiceWithCommitmentsInternal(requestId, commitments, blsPubkey);
+    }
+
+    /// @notice Per-operator cap on TEE attestation commitments at approval.
+    /// @dev Cold SSTORE per pushed entry is ~20K gas across 3 slots (~60K gas/entry).
+    ///      Without a cap, a malicious operator can submit a list large enough to
+    ///      gas-brick `_persistTeeCommitments` during the final operator's activation
+    ///      call, permanently stalling the service. 8 is well above any realistic
+    ///      number of acceptable backends an operator would commit to.
+    uint256 internal constant MAX_TEE_COMMITMENTS_PER_OPERATOR = 8;
+
+    /// @notice Maximum TTL on an operator's TEE attestation commitment.
+    /// @dev Without an upper bound, a commitment with `expiresAt = type(uint64).max`
+    ///      is effectively never-expiring, which defeats the "expiry forces
+    ///      re-attestation" intent of the field. 90 days is enough headroom for any
+    ///      realistic service lifetime; longer-lived services should re-commit on
+    ///      a renewal cadence rather than pin a single attestation indefinitely.
+    uint64 internal constant MAX_TEE_COMMITMENT_TTL = 90 days;
+
+    /// @notice Pre-flight authorization for an operator approving a request.
+    /// @dev Mirrors the auth gates inside `_approveServiceWithCommitmentsInternal`
+    ///      (request-not-rejected, operator-active, operator-in-request, not-already-approved).
+    ///      Hoisted up so storage-mutating paths in newer entrypoints can fail fast.
+    function _requireApprovingOperator(uint64 requestId) internal view {
+        Types.ServiceRequest storage req = _getServiceRequest(requestId);
+        if (req.rejected) revert Errors.ServiceRequestAlreadyProcessed(requestId);
+        if (!_staking.isOperatorActive(msg.sender)) revert Errors.OperatorNotActive(msg.sender);
+
+        bool isOperator = false;
+        address[] storage ops = _requestOperators[requestId];
+        for (uint256 i = 0; i < ops.length; i++) {
+            if (ops[i] == msg.sender) {
+                isOperator = true;
+                break;
+            }
+        }
+        if (!isOperator) revert Errors.Unauthorized();
+
+        if (_requestApprovals[requestId][msg.sender]) {
+            revert Errors.AlreadyApproved(requestId, msg.sender);
+        }
+    }
+
+    /// @notice Validate operator-supplied TEE attestation commitments.
+    /// @dev Reverts on: list too long; `Unset` or `DirectTdx` backend; wrong
+    ///      `nonceBinding` (not the request-derived value); zero expected-measurement;
+    ///      expiry in the past or further out than `MAX_TEE_COMMITMENT_TTL`. Empty
+    ///      array is allowed (operator opts out of TEE binding for this approval).
+    /// @param requestId The service request being approved — used to derive the
+    ///        canonical nonce that every commitment must carry.
+    /// @param teeCommitments Operator-supplied TEE attestation commitments.
+    function _validateTeeCommitments(
+        uint64 requestId,
+        Types.TeeAttestationCommitment[] calldata teeCommitments
+    )
+        internal
+        view
+    {
+        if (teeCommitments.length > MAX_TEE_COMMITMENTS_PER_OPERATOR) {
+            revert Errors.TooManyTeeCommitments(teeCommitments.length, MAX_TEE_COMMITMENTS_PER_OPERATOR);
+        }
+        bytes32 expectedNonce = teeNonceFor(requestId);
+        uint64 nowTs = uint64(block.timestamp);
+        uint64 maxExpiresAt = nowTs + MAX_TEE_COMMITMENT_TTL;
+        for (uint256 i = 0; i < teeCommitments.length; i++) {
+            Types.TeeBackend backend = teeCommitments[i].backend;
+            if (backend == Types.TeeBackend.Unset) revert Errors.UnsetTeeBackend();
+            if (backend == Types.TeeBackend.DirectTdx) revert Errors.DirectTdxNotPermitted();
+            // The attestation document the operator commits to must contain the
+            // request-derived nonce. The contract has no off-chain verifier so it
+            // enforces the binding directly: any other value (including zero) is
+            // rejected, eliminating cross-request replay at the source.
+            if (teeCommitments[i].nonceBinding != expectedNonce) {
+                revert Errors.InvalidNonceBinding();
+            }
+            // Zero measurement is not a real hash output. Either always-fail or
+            // always-trivially-pass under the off-chain comparator — reject the
+            // ambiguity at approval rather than discover it at provision time.
+            if (teeCommitments[i].expectedMeasurement == bytes32(0)) {
+                revert Errors.InvalidExpectedMeasurement();
+            }
+            uint64 expiresAt = teeCommitments[i].expiresAt;
+            if (expiresAt != 0) {
+                if (expiresAt <= nowTs) revert Errors.TeeCommitmentExpired(expiresAt, nowTs);
+                if (expiresAt > maxExpiresAt) revert Errors.TeeCommitmentExpiryTooFar(expiresAt, maxExpiresAt);
+            }
+        }
+    }
+
+    /// @notice Canonical TEE nonce binding for `requestId` on this chain/contract.
+    /// @dev Operators must use exactly this value as `nonceBinding` in any
+    ///      `TeeAttestationCommitment` for the request. Anyone (operator, client,
+    ///      verifier, indexer) can derive it deterministically.
+    /// @param requestId The service request ID.
+    /// @return The 32-byte nonce that uniquely binds an attestation to this
+    ///         request on this contract on this chain.
+    function teeNonceFor(uint64 requestId) public view returns (bytes32) {
+        return keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid));
+    }
+
+    /// @notice Persist validated TEE commitments and emit recording events.
+    function _storeRequestTeeCommitments(
+        uint64 requestId,
+        address operator,
+        Types.TeeAttestationCommitment[] calldata teeCommitments
+    )
+        internal
+    {
+        for (uint256 i = 0; i < teeCommitments.length; i++) {
+            _requestTeeCommitments[requestId][operator].push(teeCommitments[i]);
+            emit TeeCommitmentRecorded(
+                requestId, operator, teeCommitments[i].backend, teeCommitments[i].expectedMeasurement
+            );
+        }
+    }
+
     /// @notice Internal implementation for approving with commitments and optional BLS key
     function _approveServiceWithCommitmentsInternal(
         uint64 requestId,
@@ -242,13 +404,7 @@ abstract contract ServicesApprovals is Base {
     ///         to register a public key. Bound to chainId + verifying contract + operator
     ///         address so a PoP from one chain or operator cannot be replayed.
     function blsPopMessage(address operator, uint256[4] memory blsPubkey) public view returns (bytes memory) {
-        return abi.encode(
-            "TANGLE_BLS_POP_v1",
-            block.chainid,
-            address(this),
-            operator,
-            blsPubkey
-        );
+        return abi.encode("TANGLE_BLS_POP_v1", block.chainid, address(this), operator, blsPubkey);
     }
 
     /// @dev Reverts unless `popSignature` is a valid BLS G1 signature over `blsPopMessage`
@@ -258,7 +414,10 @@ abstract contract ServicesApprovals is Base {
         address operator,
         uint256[4] memory blsPubkey,
         uint256[2] memory popSignature
-    ) internal view {
+    )
+        internal
+        view
+    {
         if (!_isNonZeroBlsPubkey(blsPubkey)) revert Errors.InvalidBLSSignature();
 
         bool ok = BN254.verifyBls(
--- a/src/facets/tangle/TangleServicesFacet.sol
+++ b/src/facets/tangle/TangleServicesFacet.sol
@@ -16,7 +16,7 @@ contract TangleServicesFacet is ServicesApprovals, IFacetSelectors {
     using EnumerableSet for EnumerableSet.AddressSet;
 
     function selectors() external pure returns (bytes4[] memory selectorList) {
-        selectorList = new bytes4[](7);
+        selectorList = new bytes4[](10);
         selectorList[0] = this.approveService.selector;
         selectorList[1] = bytes4(keccak256("approveServiceWithCommitments(uint64,((uint8,address),uint16)[])"));
         selectorList[2] = this.rejectService.selector;
@@ -26,6 +26,33 @@ contract TangleServicesFacet is ServicesApprovals, IFacetSelectors {
         );
         selectorList[5] = this.getOperatorBlsPubkey.selector;
         selectorList[6] = this.blsPopMessage.selector;
+        selectorList[7] = bytes4(
+            keccak256(
+                "approveServiceWithTeeCommitments(uint64,((uint8,address),uint16)[],uint256[4],uint256[2],(uint8,bytes32,bytes32,uint64)[])"
+            )
+        );
+        selectorList[8] = this.getTeeCommitment.selector;
+        selectorList[9] = this.teeNonceFor.selector;
+    }
+
+    /// @notice Read the operator's TEE attestation commitments for a service.
+    /// @dev Empty array if the operator approved without TEE commitments.
+    /// @param serviceId The active service ID
+    /// @param operator The operator whose commitments to read
+    /// @return commitments Array of recorded TEE commitments (matches storage order)
+    function getTeeCommitment(
+        uint64 serviceId,
+        address operator
+    )
+        external
+        view
+        returns (Types.TeeAttestationCommitment[] memory commitments)
+    {
+        Types.TeeAttestationCommitment[] storage stored = _serviceTeeCommitments[serviceId][operator];
+        commitments = new Types.TeeAttestationCommitment[](stored.length);
+        for (uint256 i = 0; i < stored.length; i++) {
+            commitments[i] = stored[i];
+        }
     }
 
     /// @notice Get operator's BLS public key for a service
@@ -65,6 +92,9 @@ contract TangleServicesFacet is ServicesApprovals, IFacetSelectors {
 
         // Persist resource commitments from request to service (hash per operator)
         _persistResourceCommitments(serviceId, requestId);
+
+        // Persist TEE attestation commitments from request to service (per operator).
+        _persistTeeCommitments(serviceId, requestId);
 
         (uint16[] memory exposures, uint256 totalExposure) = _assignOperatorsFromRequest(serviceId, requestId);
 
@@ -298,6 +328,21 @@ contract TangleServicesFacet is ServicesApprovals, IFacetSelectors {
             // Only transfer if non-zero
             if (reqKey.key[0] != 0 || reqKey.key[1] != 0 || reqKey.key[2] != 0 || reqKey.key[3] != 0) {
                 _serviceOperatorBlsPubkeys[serviceId][op] = reqKey;
+            }
+        }
+    }
+
+    /// @notice Copy TEE attestation commitments from request to service for every operator.
+    /// @dev Skips operators that did not record commitments at approval time.
+    function _persistTeeCommitments(uint64 serviceId, uint64 requestId) private {
+        address[] storage requestOperators = _requestOperators[requestId];
+        for (uint256 i = 0; i < requestOperators.length; i++) {
+            address op = requestOperators[i];
+            Types.TeeAttestationCommitment[] storage src = _requestTeeCommitments[requestId][op];
+            uint256 len = src.length;
+            if (len == 0) continue;
+            for (uint256 j = 0; j < len; j++) {
+                _serviceTeeCommitments[serviceId][op].push(src[j]);
             }
         }
     }
--- a/src/interfaces/ITangleServices.sol
+++ b/src/interfaces/ITangleServices.sol
@@ -131,7 +131,8 @@ interface ITangleServices {
         uint8 stakingPercent,
         uint256[4] calldata blsPubkey,
         uint256[2] calldata popSignature
-    ) external;
+    )
+        external;
 
     /// @notice Approve a service request with both security commitments and BLS public key
     /// @param requestId The service request ID
@@ -145,6 +146,38 @@ interface ITangleServices {
         uint256[2] calldata popSignature
     )
         external;
+
+    /// @notice Approve a service request with security commitments, BLS pubkey, and TEE
+    ///         attestation commitments. Each commitment binds the operator to a specific
+    ///         backend + measurement that must match the live attestation off-chain.
+    /// @param requestId The service request ID
+    /// @param commitments Security commitments matching the request requirements
+    /// @param blsPubkey The operator's BLS G2 public key (zero pubkey allowed if BLS unused)
+    /// @param popSignature G1 proof-of-possession signature (only validated when blsPubkey != 0)
+    /// @param teeCommitments TEE attestation commitments to record for the caller
+    function approveServiceWithTeeCommitments(
+        uint64 requestId,
+        Types.AssetSecurityCommitment[] calldata commitments,
+        uint256[4] calldata blsPubkey,
+        uint256[2] calldata popSignature,
+        Types.TeeAttestationCommitment[] calldata teeCommitments
+    )
+        external;
+
+    /// @notice Read recorded TEE commitments for an operator on an active service.
+    function getTeeCommitment(
+        uint64 serviceId,
+        address operator
+    )
+        external
+        view
+        returns (Types.TeeAttestationCommitment[] memory);
+
+    /// @notice Canonical TEE attestation nonce binding for `requestId` on this
+    ///         contract on this chain. Operators MUST submit this exact value as
+    ///         `nonceBinding` in any `TeeAttestationCommitment` for the request —
+    ///         it eliminates cross-request attestation replay at approval time.
+    function teeNonceFor(uint64 requestId) external view returns (bytes32);
 
     /// @notice Build the canonical message an operator must sign with their BLS secret key
     ///         to register a public key. Bound to chainId + verifying contract + operator.
--- a/src/libraries/Errors.sol
+++ b/src/libraries/Errors.sol
@@ -438,4 +438,47 @@ library Errors {
 
     /// @notice Cross-chain message has already been processed (replay protection)
     error MessageAlreadyProcessed(bytes32 messageId);
+
+    // ═══════════════════════════════════════════════════════════════════════════
+    // TEE ATTESTATION COMMITMENTS
+    // ═══════════════════════════════════════════════════════════════════════════
+
+    /// @notice DirectTdx backend is not permitted for service approvals.
+    /// @dev Mirrors the policy enforced in
+    ///      ai-agent-sandbox-blueprint #50: only vendor-mediated TEE backends are accepted.
+    error DirectTdxNotPermitted();
+
+    /// @notice TEE commitment expiry is in the past.
+    error TeeCommitmentExpired(uint64 expiresAt, uint64 nowTimestamp);
+
+    /// @notice TEE commitment supplied with a zero or wrong nonce-binding.
+    /// @dev `nonceBinding` MUST equal `keccak256(abi.encode(requestId, address(this), block.chainid))`.
+    ///      That value uniquely identifies a single service request on a single chain
+    ///      and a single contract deployment, so an attestation document carrying it
+    ///      cannot be replayed against any other request. Operators that submit any
+    ///      other value (including zero) are rejected here — no off-chain freshness
+    ///      assumption is required.
+    error InvalidNonceBinding();
+
+    /// @notice TEE commitment expiry is too far in the future.
+    /// @dev Bounded so an operator cannot commit to a stale measurement effectively
+    ///      forever. The cap is `MAX_TEE_COMMITMENT_TTL` (see ServicesApprovals).
+    error TeeCommitmentExpiryTooFar(uint64 expiresAt, uint64 maxAllowed);
+
+    /// @notice TEE commitment supplied with a zero expected-measurement.
+    /// @dev A zero measurement is not a real hash output and matches no real workload;
+    ///      either always-fail or always-trivially-pass depending on the off-chain
+    ///      comparator. Reject at approval to remove the ambiguity.
+    error InvalidExpectedMeasurement();
+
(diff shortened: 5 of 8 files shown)


tangletools:
## ✅ No Blockers — `57b333d1`

**Readiness 87/100** · **Confidence 95/100** · 4 findings (1 medium, 3 low)

| | kimi | glm | aggregate |
|---|---|---|---|
| **Readiness** | 87 | 92 | **87** |
| **Correctness** | 92 | 95 | **92** |
| **Security** | 91 | 94 | **91** |
| **Testing** | 93 | 95 | **93** |
| **Architecture** | 88 | 93 | **88** |

> Read every changed file. The TEE commitment feature is well-structured: auth-before-writes ordering is correct, `nonReentrant` guard is in place, storage gap is correctly reduced by 2 for 2 new mappings, nonce replay prevention is sound, gas-griefing cap is justified and documented, and the request→service migration ordering comment is accurate. One documentation bug in Errors.sol will cause off-chain integrators to derive the wrong nonce and hit silent reverts; everything else is nits or consistent with existing codebase patterns. | Read all 8 changed files in full, verified all 23 tests pass

🟠 **MEDIUM** Errors.InvalidNonceBinding comment has wrong nonce formula — integrators will always compute the wrong value — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L455)
> [Line 455](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L455): the NatSpec states `nonceBinding MUST equal keccak256(abi.encode(requestId, address(this), block.chainid))`. The actual implementation in `ServicesApprovals.teeNonceFor` ([line 315](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L315)) is `keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid))`. The domain separator `"tangle.tee.nonce"` is omitted from the comment. Any operator SDK or off-chain tool that follows this comment verbatim will produce a different 32-byte value and receive `InvalidNonceBinding` on every submission — silently, with no indication that the fo

🟡 **LOW** Double O(n) operator-list scan per approveServiceWithTeeCommitments call — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L208)
> [Lines 208](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L208) and 349: `_requireApprovingOperator` iterates `_requestOperators[requestId]` to confirm `msg.sender` is in the list, then `_approveServiceWithCommitmentsInternal` (called at [line 218](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L218)) repeats the identical loop at [line 349](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L349). Under the `nonReentrant` guard no state can change between the two calls, so the second scan is pure waste — linear in operator count per approval. Existing entrypoints (`approveServ

🟡 **LOW** _requestTeeCommitments never deleted after activation — orphaned storage slots — [`src/facets/tangle/TangleServicesFacet.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L337-L348)
> [Lines 337-348](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L337-L348): `_persistTeeCommitments` copies commitment arrays from `_requestTeeCommitments[requestId][op]` to `_serviceTeeCommitments[serviceId][op]` but never deletes the source. At the documented ~60K gas per entry and up to 8 entries × N operators, the orphaned slots are non-trivial. This is consistent with `_persistResourceCommitments` ([line 294-311](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L294-L311)) which also leaves `_requestResourceRequirements` intact, so it matches the existing design choice — but both forfeit SSTORE refunds. Worth a one-line `d

🟡 **LOW** Errors.InvalidNonceBinding @dev comment omits domain prefix — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L454-L456)
> [Line 454-456](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L454-L456): The @dev comment says `nonceBinding MUST equal keccak256(abi.encode(requestId, address(this), block.chainid))` but the actual implementation in ServicesApprovals.sol:314-316 uses `keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid))`. An integrator reading the error doc to compute the canonical nonce would derive the wrong value and get rejected. Fix: add the 'tangle.tee.nonce' prefix to the @dev comment.

---
<sub>tangletools · 2026-05-04T20:12:34Z · [trace](https://gist.github.com/drewstone/a0c7865f8dc1b74b24e8b378d94ae1cc)</sub>

tangletools (dismissed):
## ✅ Approved — 5 non-blocking findings

> The AppRegistry and TEE attestation commitment features (6 of 14 changed files) are clean, well-tested, and architecturally sound. However this PR also bundles four serious regressions in the governance, rewards, and streaming-payment layers: the TangleTimelock._setMinDelay function now writes to hardcoded slot 0x33 instead of the correct ERC-7201 namespaced slot, meaning updateDelay silently fails to persist; the updateDelay auth model was downgraded from self-call to DEFAULT_ADMIN_ROLE (governance bypass); nonReentrant was removed from two load-bearing streaming-payment drip functions; and I

🟠 **MEDIUM** TeeAttestationCommitment.nonceBinding not validated — anti-replay protection is opt-in — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/core/ServicesApprovals.sol#L231-L248)
> [Lines 231-248](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/core/ServicesApprovals.sol#L231-L248) (`_validateTeeCommitments`): validates `backend != DirectTdx` and `expiresAt > now`, but does not check `nonceBinding != bytes32(0)`. The `Types.TeeAttestationCommitment` NatSpec (Types.sol) documents: 'Caller-supplied nonce that binds this commitment to a specific challenge ... so a stale attestation cannot be replayed against a fresh request.' If an operator supplies `nonceBinding = 0`, the commitment has no challenge binding and a recycled attestation from an old service can satisfy it. Add: `if (teeCommitments[i].nonceBinding == bytes32(0)) revert Errors.InvalidNonceBinding()`.

🟠 **MEDIUM** AppRegistry.listApps returns revoked apps — callers must do a second lookup to filter — [`src/registry/AppRegistry.sol`](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L357)
> [Line 357](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L357) (`revokeApp`): sets `record.revoked = true` but does not remove `appId` from `_namespaceApps[namespace]`. `listApps` ([line 368](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L368)) returns `_namespaceApps[namespace].values()` which includes revoked apps. Any consumer iterating `listApps` results and treating them as active apps will silently include revoked entries. Fix: either remove the appId from `_namespaceApps` in `revokeApp`, or add a `listActiveApps` view that filters on `revoked`.

🟡 **LOW** Two error types defined but never used in changed code — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/libraries/Errors.sol#L455)
> [Lines 455](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/libraries/Errors.sol#L455) and 459: MeasurementMismatch and TeeCommitmentLengthMismatch are declared in the new TEE section but are never referenced by any contract in this PR. The MeasurementMismatch comment says 'Reserved for off-chain / blueprint use' which is fine for forward-compatibility, but TeeCommitmentLengthMismatch has no apparent caller at all. If these are genuinely needed by off-chain code (e.g. replica verification in a blueprint), that should be documented — otherwise they're dead code adding to the ABI/bytecode.

🟡 **LOW** AppRecord.verified is stale cached storage — getApp bypasses it anyway — [`src/registry/AppRegistry.sol`](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L207-L218)
> [Lines 207-218](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L207-L218) (`publishApp`) and 237 (`pinAppVersion`): `record.verified = _namespaceVerified[namespace]` writes the verified flag into `AppRecord` storage on every publish and pin. But `getApp` ([line 302](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L302)) returns `verified = _namespaceVerified[namespace]`, reading directly from the source mapping and ignoring `record.verified`. The cached field is written on every publish/pin (~5k gas SSTORE), never read, and is stale between pins. Remove the `record.verified` field and its writes to save gas and eliminate the misleading cached state.

🟡 **LOW** appCount uses EnumerableSet.length() which iterates O(n) — [`src/registry/AppRegistry.sol`](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L375-L376)
> [Line 375-376](https://github.com/tangle-network/tnt-core/blob/cbdf527f1df19b7d7add6107d802521b856f2f21/src/registry/AppRegistry.sol#L375-L376): appCount returns _namespaceApps[namespace].length() which iterates the entire EnumerableSet to count elements. If a namespace accumulates many apps, this becomes expensive. A separate counter mapping (incremented on publish, decremented on any future removal) would provide O(1). Not a blocker since there's no app removal path yet, but worth noting for future scale.

---
<sub>tangletools · 2026-05-04T18:20:59Z · [trace](https://gist.github.com/drewstone/5e8e7ff002130554dfb22ff5b2da880d)</sub>

tangletools (approved):
## ✅ Approved — 4 non-blocking findings

> Read all 7 changed files plus the full ServicesApprovals.sol and TangleServicesFacet._activateService flow. The core logic is correct: TEE commitments are stored to _requestTeeCommitments before _approveServiceWithCommitmentsInternal triggers activation, which copies them into _serviceTeeCommitments via _persistTeeCommitments — the ordering is intentional and correct. Storage gap math (44→42 for 2 new mapping slots) is right. ABI selector keccak string matches the canonical tuple encoding. The one structural concern is that _storeRequestTeeCommitments performs SSTOREs before the operator-autho

🟠 **MEDIUM** Authorization check follows storage write — breaks validate-before-write — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/11e89d49d5af671df58cde83cbd674082c8b91bc/src/core/ServicesApprovals.sol#L208-L214)
> In approveServiceWithTeeCommitments ([lines 208–214](https://github.com/tangle-network/tnt-core/blob/11e89d49d5af671df58cde83cbd674082c8b91bc/src/core/ServicesApprovals.sol#L208-L214)), _storeRequestTeeCommitments (which does SSTORE pushes) is called before _approveServiceWithCommitmentsInternal, which is where the operator-in-request-list check and AlreadyApproved guard live. An unauthorized or already-approved caller executes the storage loop before the transaction reverts; the writes are rolled back, so there is no persistent corruption, but the caller burns gas on the loop and the pattern is wrong. Fix: move _validateTeeCommitments + BLS check before the storage write, and restructure so authorization is confirmed before any state mutation.

🟡 **LOW** _requestTeeCommitments not reclaimed after activation or rejection — [`src/facets/tangle/TangleServicesFacet.sol`](https://github.com/tangle-network/tnt-core/blob/11e89d49d5af671df58cde83cbd674082c8b91bc/src/facets/tangle/TangleServicesFacet.sol)
> _persistTeeCommitments (line ~333–347) copies from _requestTeeCommitments to _serviceTeeCommitments but never deletes the source. After activation the request-scoped storage is never accessed again (comment in TangleStorage.sol confirms this) but the slots remain warm, costing refund gas that is never collected. The same pattern exists for other request-scoped maps (consistent), but note that deleting inside _persistTeeCommitments after the copy would reclaim ~4700 gas per non-zero slot. Low priority given consistency with the rest of the codebase.

🟡 **LOW** TeeCommitmentLengthMismatch error declared but never used — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/11e89d49d5af671df58cde83cbd674082c8b91bc/src/libraries/Errors.sol)
> TeeCommitmentLengthMismatch(uint256 commitmentCount, uint256 teeCommitmentCount) is declared at the bottom of the TEE section but is referenced nowhere in this PR (no call site in ServicesApprovals.sol or TangleServicesFacet.sol). If the per-operator commitment-count enforcement it was intended to guard was dropped, delete the error. If the enforcement is still planned, add the check.

🟡 **LOW** No test for non-TEE approval path producing empty getTeeCommitment result — [`test/tangle/TeeCommitmentApprovalTest.t.sol`](https://github.com/tangle-network/tnt-core/blob/11e89d49d5af671df58cde83cbd674082c8b91bc/test/tangle/TeeCommitmentApprovalTest.t.sol)
> All tests use approveServiceWithTeeCommitments. There is no test that verifies an operator who approved via approveService, approveServiceWithBls, or approveServiceWithCommitmentsAndBls results in getTeeCommitment returning an empty array after activation. This exercises the 'if (len == 0) continue' branch in _persistTeeCommitments for the non-TEE path and confirms the storage separation is correct.

---
<sub>tangletools · 2026-05-04T18:45:32Z · [trace](https://gist.github.com/drewstone/53138bf6979a3d2ef32161d8f5fdeaba)</sub>

tangletools (dismissed):
## ❌ 1 Blocking Finding

> The TEE commitment feature is architecturally clean — correct storage layout, proper ordering of TEE writes before the activation-triggering approval call, correct storage gap reduction (44→42 for two new mappings), and a well-structured test suite that covers the primary error paths. One high-severity issue undermines the rating: there is no upper bound on the `teeCommitments` calldata array, which lets a malicious operator in a multi-operator service submit a gas-exhausting array that permanently bricks service activation when the last legitimate operator triggers it. Two medium gaps round o

🔴 **HIGH** Unbounded teeCommitments array enables permanent activation DoS in multi-operator services — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L208-L214)
> [Lines 208–214](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L208-L214) and 239–252: `_validateTeeCommitments` and `_storeRequestTeeCommitments` iterate over `teeCommitments` with no length cap. Each entry pushes to `_requestTeeCommitments[requestId][operator]` (cold SSTORE ~20K gas per slot; struct is 3 slots → ~60K gas/entry). These stored entries are later bulk-copied in `_persistTeeCommitments` (TangleServicesFacet.sol:336–347), which runs inside `_activateService` triggered by whoever submits the *final* approval. A malicious operator in a 2+ operator service can submit ~400 entries in their own approval tx (staying under block gas limit), then the final legitimate operator's approval will hit the gas limit dur

### 4 Non-Blocking

🟠 **MEDIUM** Zero expectedMeasurement silently accepted — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L220-L236)
> [Lines 220–236](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L220-L236): `_validateTeeCommitments` rejects `nonceBinding == bytes32(0)` ([line 228](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L228)) with `InvalidNonceBinding`, but does not reject `expectedMeasurement == bytes32(0)`. An operator submitting a zero measurement commits to a workload whose measurement is zero — practically impossible for any real hash function output, but semantically a commitment that can never be satisfied. An off-chain provisioning oracle comparing `attestedMeasurement == expectedMeasurement` against a zero value would either always fail (stalling the service)

🟠 **MEDIUM** No test for mixed TEE / non-TEE approvals in a multi-operator service — [`test/tangle/TeeCommitmentApprovalTest.t.sol`](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/test/tangle/TeeCommitmentApprovalTest.t.sol#L185)
> All multi-operator test cases (e.g. `test_multipleOperators_distinctBackends_persistedPerOperator`, [line 185](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/test/tangle/TeeCommitmentApprovalTest.t.sol#L185)) send every operator through `approveServiceWithTeeCommitments`. The `_persistTeeCommitments` function in TangleServicesFacet.sol:336–347 has explicit logic to skip operators with no stored commitments (`if (len == 0) continue;`), but this branch is never exercised. The production scenario where operator 1 uses `approveServiceWithTeeCommitments` and operator 2 uses plain `approveServiceWithCommitments` is unverified; a bug in the skip logic or in how activation counts operators would not be caught.

🟡 **LOW** Duplicate TeeBackend entries silently accepted — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L220-L236)
> [Lines 220–236](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/core/ServicesApprovals.sol#L220-L236): `_validateTeeCommitments` does not check for duplicate `backend` values across entries. An operator can submit two entries with `backend == TeeBackend.AwsNitro` and different (or identical) measurements; both are stored and two `TeeCommitmentRecorded` events are emitted. The NatSpec says 'any one matching satisfies the policy,' which is consistent with duplicates being harmless at the protocol level, but they waste storage and inflate event logs. If the intent is per-backend commitment, add a seen-bitmap check during validation.

🟡 **LOW** TeeBackend enum default is Phala (0), not caught by validation — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/028db05d5a2f30e98656cdced63760fc7f89ddff/src/libraries/Types.sol)
> The TeeBackend enum starts at Phala=0, so a struct where `backend` is accidentally left unset defaults to Phala. The `nonceBinding==0` check at src/core/ServicesApprovals.sol:228 catches the all-zero case, but if an integrator sets nonceBinding and expectedMeasurement while forgetting backend, it silently accepts Phala. Not exploitable in practice but a footgun. Consider starting the enum at 1 with a sentinel value like `Unset`.

---
<sub>tangletools · 2026-05-04T19:04:47Z · [trace](https://gist.github.com/drewstone/63de10c6f74d44bc20bf8dfba13c78c8)</sub>

drewstone:
Addressed all five findings from the latest review (commit `bce6788`):

### 🔴 HIGH — Unbounded `teeCommitments` DoS

Added `MAX_TEE_COMMITMENTS_PER_OPERATOR = 8` constant and length check in `_validateTeeCommitments`. Above the cap reverts with `Errors.TooManyTeeCommitments(supplied, max)`. 8 is well above any realistic number of acceptable backends an operator would commit to (5 backends in the enum minus DirectTdx, leaving 4 viable), and well below the gas-bricking threshold.

### 🟠 MEDIUM — Authorization check follows storage write

Hoisted the auth gates (request-not-rejected, operator-active, operator-in-request, not-already-approved) into a new `_requireApprovingOperator(requestId)` internal view, called BEFORE `_validateTeeCommitments` and BEFORE `_storeRequestTeeCommitments`. The internal `_approveServiceWithCommitmentsInternal` retains its own auth check (defensive consistency), so the cost is one extra read of `req.rejected` + the operator-list scan — paid only on the happy path.

### 🟠 MEDIUM — Zero `expectedMeasurement` silently accepted

Added `Errors.InvalidExpectedMeasurement` and reject in `_validateTeeCommitments`. Same treatment as the existing zero-`nonceBinding` check.

### 🟠 MEDIUM — No test for mixed TEE / non-TEE multi-operator service

Added `test_mixedTeeAndNonTee_acrossOperators_persistsCorrectly`: operator1 approves with `approveServiceWithTeeCommitments`, operator2 + operator3 use plain `approveServiceWithCommitments`. Service activates on the third approval. Asserts operator1's commitment array persists onto the service and operators 2/3 have empty arrays — exercising the `if (len == 0) continue;` skip branch in `_persistTeeCommitments`.

### 🟡 LOW — `TeeBackend` enum default is Phala (0)

Inserted `Unset` at index 0 of the enum + `Errors.UnsetTeeBackend` + reject in `_validateTeeCommitments`. An integrator who forgets to populate the backend field now hits `UnsetTeeBackend` instead of accidentally committing to whichever real backend was at index 0.

### 🟡 LOW — Duplicate `TeeBackend` entries (deferred)

Not addressed — the NatSpec on the entrypoint is explicit that "any one matching at provisioning time satisfies the policy", so duplicates are semantically harmless. With the per-operator cap of 8, the storage waste from a duplicate is bounded. If the off-chain provisioning oracle needs uniqueness, it can dedup before submitting; making it a contract-level invariant would muddy the documented semantics.

### Tests

10 → 16 cases. New: `test_approve_zeroExpectedMeasurement_reverts`, `test_approve_unsetBackend_reverts`, `test_approve_tooManyCommitments_reverts`, `test_approve_atCommitmentCap_succeeds`, `test_approve_unauthorizedCaller_revertsBeforeStorage`, `test_mixedTeeAndNonTee_acrossOperators_persistsCorrectly`.

Force-pushed to keep the branch a single commit.

tangletools (approved):
## ✅ Approved — 4 non-blocking findings

> Read all 7 changed files plus Base.sol/_getServiceRequest, TangleServicesFacet._activateService, and the full ServicesApprovals context. No critical or high bugs found. Storage gap math is correct (2 new mappings → gap reduced 44→42). Auth-first ordering is sound, DoS cap is justified, and all six validation rules in _validateTeeCommitments are semantically correct. Two medium/low findings: a double-auth-check in the new entrypoint wastes ~8 K gas on the happy path, and _persistTeeCommitments never deletes _requestTeeCommitments source slots after copying (forfeits storage refund). One enum ba

🟠 **MEDIUM** Double auth check in approveServiceWithTeeCommitments wastes gas — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/core/ServicesApprovals.sol#L208)
> approveServiceWithTeeCommitments ([line 208](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/core/ServicesApprovals.sol#L208)) calls _requireApprovingOperator(requestId) which reads four storage slots (req.rejected, isOperatorActive, _requestOperators[requestId], _requestApprovals[requestId][msg.sender]), then calls _approveServiceWithCommitmentsInternal ([line 218](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/core/ServicesApprovals.sol#L218)) which reads the same four slots again ([lines 309-325](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/core/ServicesApprovals.sol#L309-L325)). The only state change between the two reads is the TEE commitment write

🟡 **LOW** TEE section in TangleStorage.sol lacks slot-range comment — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/TangleStorage.sol#L372-L383)
> Every other storage section in TangleStorage.sol carries a slot-range annotation (e.g. '// PRICING STORAGE (Slot 136-140)'). The new TEE ATTESTATION COMMITMENTS section ([lines 372-383](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/TangleStorage.sol#L372-L383)) has no such annotation, making it harder to track the cumulative slot count when adding future storage. Add the slot range to match the surrounding convention.

🟡 **LOW** _persistTeeCommitments does not delete source slots — no storage refund — [`src/facets/tangle/TangleServicesFacet.sol`](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/facets/tangle/TangleServicesFacet.sol#L334)
> _persistTeeCommitments ([line 334](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/src/facets/tangle/TangleServicesFacet.sol#L334)) copies each TeeAttestationCommitment element from _requestTeeCommitments[requestId][op] into _serviceTeeCommitments[serviceId][op] but never calls delete or pop on the source array. Each struct occupies 4 storage slots (backend/uint8, expectedMeasurement/bytes32, nonceBinding/bytes32, expiresAt/uint64). With up to 8 elements per operator and N operators, the request-scoped data (up to 32×N slots) is permanently wasted. This is consistent with the existing _persistServiceSecurity pattern, but with TEE commitments there are potentially more slots per operator. Consider deleting the source array after copy or documenting t

🟡 **LOW** AzureSkr backend has zero test coverage — [`test/tangle/TeeCommitmentApprovalTest.t.sol`](https://github.com/tangle-network/tnt-core/blob/bce6788faea09bd500f33f7221b7d622cdf92052/test/tangle/TeeCommitmentApprovalTest.t.sol)
> Types.TeeBackend.AzureSkr (enum value 4) is defined in Types.sol (line ~305) and is accepted by _validateTeeCommitments (not Unset, not DirectTdx), but no test in TeeCommitmentApprovalTest exercises it. Phala, AwsNitro, GcpConfidential, and DirectTdx are all covered. Add a test mirroring test_approve_withNitroCommitment_persistsOnActivation but using TeeBackend.AzureSkr to confirm it round-trips correctly through storage.

---
<sub>tangletools · 2026-05-04T19:53:32Z · [trace](https://gist.github.com/drewstone/889af2b868eb4d80f12dfc249c6006cf)</sub>

drewstone:
Adversarial sweep on top of the review-fix commit (`57b333d`). Two REAL findings remediated, one documented with measurement.

### T1 [HIGH] — `nonceBinding` was operator-supplied, not bound to anything on-chain → REPLAY POSSIBLE

**Before this push:** validation only checked `nonceBinding != bytes32(0)`. An operator could submit `nonceBinding = N` for request A, produce one attestation document with `nonce = N`, then resubmit `nonceBinding = N` for request B — same attestation satisfies both.

**Why it slipped past prior review:** the NatSpec said "binds this commitment to a specific challenge so a stale attestation cannot be replayed against a fresh request." But the contract enforced none of that — the binding semantics were entirely off-chain.

**Fix.** `nonceBinding` MUST equal `keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid))`. New public view `teeNonceFor(uint64) → bytes32` exposes the canonical value. Cross-request replay becomes structurally impossible — an attestation document binding to nonce of request A literally cannot satisfy a commitment requiring nonce of request B.

### T4 [LOW-MEDIUM] — `expiresAt` had no upper bound → silent never-expiry

**Before:** `expiresAt = type(uint64).max` was accepted (~year 584,554,051,223). A commitment with that expiry was effectively never-expiring, defeating the "force re-attestation" intent.

**Fix.** `MAX_TEE_COMMITMENT_TTL = 90 days`. Validation rejects `expiresAt > block.timestamp + MAX_TEE_COMMITMENT_TTL` with `Errors.TeeCommitmentExpiryTooFar(supplied, maxAllowed)`. `expiresAt = 0` (the documented no-expiry sentinel) remains intentional.

### T2 [HIGH/system-design] — Activation gas measured, not "fixed"

`_persistTeeCommitments` runs in the activator's tx and copies every operator's commitment array onto the service. Linear in `operator_count × commitments_per_operator`. With 3 operators × 8 commits, **2,886,434 gas** for TEE persistence alone. At 10 operators × 8 → ~9.6M. At 20 × 8 → ~19M, near block-gas territory.

This is a **system-design bound**, not a contract bug. Mitigations span beyond the TEE feature (operator-count cap on TEE-required services, separate operator-paid claim step, or moving persistence to event-log lookup). New test `test_T2_activation_gasScalesWithOperatorsTimesCap` emits the number to forge log so any regression is visible.

### Adversarial coverage added

`test/tangle/TeeCommitmentHardenTest.t.sol` — 7 new tests, each with a clear PASS/FAIL semantic:

```
test_T1_replayAcrossRequests_rejected           — nonce A submitted to request B reverts
test_T1_arbitraryNonce_rejected                  — fabricated nonce reverts
test_T1_teeNonceFor_uniquePerRequest             — distinct/stable across requests
test_T2_activation_gasScalesWithOperatorsTimesCap — emits 2.89M gas measurement
test_T4_uint64Max_rejected                        — far-future reverts
test_T4_oneDayPastCap_rejected                    — 91-day reverts
test_T4_exactlyAtCap_accepted_oneSecondPast_rejected — boundary test
```

### Final test surface

- 16 happy-path / review-fix tests in `TeeCommitmentApprovalTest.t.sol` — all green
- 7 adversarial / hardening tests in `TeeCommitmentHardenTest.t.sol` — all green
- 23/23 forge tests pass locally (`forge test --match-contract "TeeCommitment(Approval|Harden)"`)

Changes added to the existing single commit; force-pushed.

drewstone:
## CI status — pre-existing OOM, not caused by this PR

[Run 25340146909](https://github.com/tangle-network/tnt-core/actions/runs/25340146909) failed at the **\"Run Forge build\"** step with **exit code 143** (SIGTERM, OOM-kill) after 17m17s of compilation. The job was killed mid-compile of 342 Solidity files; no test ever ran.

This is the same failure mode as:

- [Run 25336884245](https://github.com/tangle-network/tnt-core/actions/runs/25336884245) — same branch's prior commit, exit 143 at the build step
- [Run 25332144175](https://github.com/tangle-network/tnt-core/actions/runs/25332144175) — main, immediately after PR #116 merge, exit 143
- [Run 25330677940](https://github.com/tangle-network/tnt-core/actions/runs/25330677940) — main, immediately after PR #115 merge, exit 143

**Cause:** `via_ir = true` on `[profile.default]` in `foundry.toml` (inherited by `profile.ci`). The IR pipeline is memory-heavy on a 342-file codebase; the github-hosted `ubuntu-latest` runner (16 GB RAM) doesn't fit it. Both #115 and #116 landed despite this failing CI.

**Local reproduction is green:**

```
forge test --match-contract "TeeCommitment(Approval|Harden)" -vv
[PASS]  16 happy/review-fix tests       — TeeCommitmentApprovalTest.t.sol
[PASS]   7 adversarial/hardening tests   — TeeCommitmentHardenTest.t.sol
Suite result: ok. 23 passed; 0 failed; 0 skipped
```

This PR does not change `foundry.toml`, the CI workflow, or anything that would alter compile-step memory usage. The CI fix belongs in a separate PR scoped to the build infra (options: switch `profile.ci` to `via_ir = false`, drop `--sizes` from the build, or move to a larger runner). Happy to open that as a follow-up if you want, but not bundling it here.

The TEE commitment code itself + the harden fixes are verified locally and via PR review.

tangletools (approved):
## ✅ Approved — 4 non-blocking findings

> Read every changed file. The TEE commitment feature is well-structured: auth-before-writes ordering is correct, `nonReentrant` guard is in place, storage gap is correctly reduced by 2 for 2 new mappings, nonce replay prevention is sound, gas-griefing cap is justified and documented, and the request→service migration ordering comment is accurate. One documentation bug in Errors.sol will cause off-chain integrators to derive the wrong nonce and hit silent reverts; everything else is nits or consistent with existing codebase patterns. | Read all 8 changed files in full, verified all 23 tests pass

🟠 **MEDIUM** Errors.InvalidNonceBinding comment has wrong nonce formula — integrators will always compute the wrong value — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L455)
> [Line 455](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L455): the NatSpec states `nonceBinding MUST equal keccak256(abi.encode(requestId, address(this), block.chainid))`. The actual implementation in `ServicesApprovals.teeNonceFor` ([line 315](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L315)) is `keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid))`. The domain separator `"tangle.tee.nonce"` is omitted from the comment. Any operator SDK or off-chain tool that follows this comment verbatim will produce a different 32-byte value and receive `InvalidNonceBinding` on every submission — silently, with no indication that the fo

🟡 **LOW** Double O(n) operator-list scan per approveServiceWithTeeCommitments call — [`src/core/ServicesApprovals.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L208)
> [Lines 208](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L208) and 349: `_requireApprovingOperator` iterates `_requestOperators[requestId]` to confirm `msg.sender` is in the list, then `_approveServiceWithCommitmentsInternal` (called at [line 218](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L218)) repeats the identical loop at [line 349](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/core/ServicesApprovals.sol#L349). Under the `nonReentrant` guard no state can change between the two calls, so the second scan is pure waste — linear in operator count per approval. Existing entrypoints (`approveServ

🟡 **LOW** _requestTeeCommitments never deleted after activation — orphaned storage slots — [`src/facets/tangle/TangleServicesFacet.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L337-L348)
> [Lines 337-348](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L337-L348): `_persistTeeCommitments` copies commitment arrays from `_requestTeeCommitments[requestId][op]` to `_serviceTeeCommitments[serviceId][op]` but never deletes the source. At the documented ~60K gas per entry and up to 8 entries × N operators, the orphaned slots are non-trivial. This is consistent with `_persistResourceCommitments` ([line 294-311](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/facets/tangle/TangleServicesFacet.sol#L294-L311)) which also leaves `_requestResourceRequirements` intact, so it matches the existing design choice — but both forfeit SSTORE refunds. Worth a one-line `d

🟡 **LOW** Errors.InvalidNonceBinding @dev comment omits domain prefix — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L454-L456)
> [Line 454-456](https://github.com/tangle-network/tnt-core/blob/57b333d177d646d7896071cefb56917f37c1c3fb/src/libraries/Errors.sol#L454-L456): The @dev comment says `nonceBinding MUST equal keccak256(abi.encode(requestId, address(this), block.chainid))` but the actual implementation in ServicesApprovals.sol:314-316 uses `keccak256(abi.encode("tangle.tee.nonce", requestId, address(this), block.chainid))`. An integrator reading the error doc to compute the canonical nonce would derive the wrong value and get rejected. Fix: add the 'tangle.tee.nonce' prefix to the @dev comment.

---
<sub>tangletools · 2026-05-04T20:12:34Z · [trace](https://gist.github.com/drewstone/a0c7865f8dc1b74b24e8b378d94ae1cc)</sub>

PR #147, opened 2026-05-17
chore(v0.17.1): land stranded M-2 + H-1 + F-001 + operator-cap polish + bindings v0.17.1

drewstone:
Rescue PR — relands the M-2/H-1/F-001 audit follow-up + the polish (configurable operator cap, pull-payment refactor, audit-tag sweep) + the regenerated bindings v0.17.1. **All of this content was stranded after the stacked merge of #145 + #146 only landed in the stacked-PR base branches, not on `main`.** Only #144 actually reached main.

## What's in this PR

Five cherry-picked commits, applied in order on top of `main`:

| Commit | Summary |
|---|---|
| **M-2** | `withdrawRemainingEscrowTo(serviceId, to)` — owner-chosen recipient escape hatch when the escrow token blocklists the service owner. |
| **H-1** | Per-(serviceId, op, asset) USD price snapshot captured at activation/join, reused at every bill. Post-activation oracle drift can no longer inflate one operator's bill share. |
| **F-001 (atomic)** | First-pass atomic ERC20 staker-share forward — landed into the stacked branch but superseded by the polish commit below. |
| **Polish** | (a) `MAX_OPERATORS_PER_SERVICE` is no longer a hardcoded 64 — replaced by governance-tunable `_maxOperatorsPerService`, default **256**, admin setter. (b) F-001 refactored from atomic-wrapper plumbing to standard **pull-payment**: distributor pulls via `safeTransferFrom`, caller revokes residual allowance on revert. Net code REMOVAL (-203 / +151 across 50 files). (c) Audit-tag sweep across `src/` + `test/`. |
| **Bindings v0.17.1** | `cargo xtask gen-bindings` regenerated against the polish HEAD. ABI/struct changes: new selectors (`withdrawRemainingEscrowTo`, `setMaxOperatorsPerService`, `maxOperatorsPerService`), new events (`MaxOperatorsPerServiceUpdated`, `PushTransferFailed`, `PriceOracleFallback`), updated distributor pull-payment shape. |

## Test plan

- [x] `FOUNDRY_PROFILE=local_build forge build` clean (post-cherry-pick)
- [x] `FOUNDRY_PROFILE=fast forge test --no-match-path test/Integration.t.sol --no-match-test invariant` — **1452/1452 green** (verified on the polish branch which has the same content)
- [x] Bindings regen completed (1.1s warm-cache rebuild)

## Why this PR exists

[...]

PR #149, opened 2026-05-17
fix(payments): split refund out of TanglePaymentsFacet to clear EIP-170

tangletools:
## Summary

`TanglePaymentsFacet` runtime bytecode reached **24874 bytes (298 over the 24576-byte EIP-170 limit)** after PR #147 added `withdrawRemainingEscrowTo`. Local anvil hides this because it runs with `--disable-code-size-limit`, but Base Sepolia and every other production chain enforces the limit strictly — the facet would not deploy.

Extract the post-termination refund path into a new `PaymentsRefund` mixin and route those two selectors via `TanglePaymentsRewardsFacet`, which already owns the rewards/admin/views surface and had 8.8 KB of headroom.

## Sizes (default profile, optimizer on)

| Facet | Before | After | Δ | Headroom |
|---|---:|---:|---:|---:|
| **TanglePaymentsFacet** | 24874 | **24104** | **−770** | **472** |
| TanglePaymentsRewardsFacet | 15751 | 16280 | +529 | 8296 |
| TanglePaymentsDistributionFacet | 19651 | 19651 | 0 | 4925 |

Net facet bytecode: −241 B.

## Public surface

Unchanged. The diamond router wires selectors via each facet's `selectors()` view, so `withdrawRemainingEscrow{,To}` still dispatch through the same calldata path on `Tangle.sol`, just to a different facet. `ITangleServices.sol` interface is not modified.

## Test plan

- [x] `Payments.t.sol::WithdrawRemainingEscrowTo*` — 3/3 pass (covers new routing, `ZeroAddress` revert, `NotServiceOwner` gate)
- [x] `EndToEndSubscriptionTest.t.sol` — 9/9 pass (full subscription lifecycle including post-termination refund)
- [x] `SubscriptionEscrowInvariant.t.sol` — 3/3 pass (3,837 `refundRemaining` handler invocations against the new routing)
- [ ] Full `forge test` (running locally; will report)
- [ ] CI green

## Follow-ups

Adding a CI gate that builds with the default profile and asserts every facet ≤ 24576 — anvil's `--disable-code-size-limit` masks regressions locally, and this PR is what we needed it to catch in the first place. Tracking separately so this fix can land first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3351,7 +3351,7 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tnt-core-bindings"
-version = "0.17.0"
+version = "0.17.1"
 dependencies = [
  "alloy",
  "alloy-contract",
@@ -3361,7 +3361,7 @@ dependencies = [
 
 [[package]]
 name = "tnt-core-fixtures"
-version = "0.17.0"
+version = "0.17.1"
 
 [[package]]
 name = "tokio"
--- a/src/core/PaymentsEscrow.sol
+++ b/src/core/PaymentsEscrow.sol
@@ -7,7 +7,9 @@ import { Errors } from "../libraries/Errors.sol";
 import { PaymentLib } from "../libraries/PaymentLib.sol";
 
 /// @title PaymentsEscrow
-/// @notice Customer-facing escrow funding and post-termination withdrawal.
+/// @notice Customer-facing escrow funding for subscription services.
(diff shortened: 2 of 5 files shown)


tangletools:
## 🔍 Reviewing `a62c3049`

| Pass | Status | ETA |
|---|---|---|
| Kimi Code K2.6 | Running (4 min) | ~5-15 min |
| opencode GLM 5.1 | Running (4 min) | ~5-15 min |

> Agent review running. Reads the actual code. This comment updates in place.

<sub>tangletools · tangle-network/tnt-core#149 · model: `kimi-for-coding` · started 2026-05-17T22:38:46Z</sub>

PR #157, opened 2026-05-26
feat(deploy): record every blueprint variant + per-binary publish

drewstone:
The multi-blueprint repos register their full set on-chain, but the manifest only captured one id each.

**On-chain truth:** `blueprintCount() = 17`. ai-agent-sandbox registered its trio (ids **10** sandbox / **11** instance / **12** tee-instance); ai-trading its quad (**13** trading / **14** instance / **15** tee-instance / **16** validator). All owned by the deployer — verified via `getBlueprint`.

The sweep's parser used `tail -1`, so `blueprints.tsv` recorded 1 of 3–4 per repo and even **mislabelled trading** (it stored id 16 = the *validator* as the main blueprint; the real `trading-blueprint` is id 13).

## Changes
- **`register-blueprints.sh`** — write one manifest row per emitted `DEPLOY_<LABEL>BLUEPRINT_ID`, pairing each with its same-label BSM. Re-runs now record every variant.
- **`blueprints.tsv`** — rebuilt as the complete 17-row manifest from on-chain data, with the operator **binary name + variant** per row (the bin→id map the publish step needs).
- **`publish-blueprint-binary.yml`** — optional `binary_name` input so a multi-blueprint repo publishes the right binary to the right id.

## Net
Registration was already complete for all 17 (the per-repo scripts deploy all variants). This makes the **manifest** complete + correct and lets the publish workflow target each variant — so once `BLUEPRINT_OWNER_KEY` is set, all 17 can be published (10 singles + 3 sandbox + 4 trading), each to its own id.

## Test plan
- [x] `bash -n register-blueprints.sh` + workflow YAML parse clean
- [x] manifest = 17 rows, ids 0–16, BSMs from on-chain

### DIFF
--- a/.github/workflows/publish-blueprint-binary.yml
+++ b/.github/workflows/publish-blueprint-binary.yml
@@ -22,6 +22,10 @@ on:
       blueprint_id:
         description: 'On-chain blueprint id (see blueprints.tsv)'
         required: true
+      binary_name:
+        description: 'Binary to publish (multi-blueprint repos, e.g. trading-instance-blueprint). Leave blank for single-binary repos.'
+        required: false
+        default: ''
       network:
         description: 'Deployment network (matches deployments/<network>/latest.json)'
         required: true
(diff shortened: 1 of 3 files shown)


tangletools (dismissed):
Records all 17 variants from on-chain truth; sweep now multi-blueprint aware; publish workflow per-binary. Approving.

PR #162, opened 2026-05-30
feat(blueprints): setBlueprintSources — owner can repoint a blueprint's binary sources

tangletools:
## Summary

`createBlueprint` writes a blueprint's `BlueprintSource[]` once at genesis, and nothing could change it afterward — `updateBlueprint` only edits metadata. This adds `setBlueprintSources(uint64, BlueprintSource[])` (owner-gated) so an owner can repoint a blueprint at corrected, fetchable, multi-arch artifacts.

Affected flow: **operator cold-start through `blueprint-manager`**. The manager resolves the *initial* operator binary from these on-chain `BlueprintSource`s — the published/active `BinaryVersion` set only drives *in-place upgrades*. So a blueprint registered with a placeholder or single-arch source is permanently un-runnable through the manager. This is exactly the live state of the trading blueprint (id 13): its source is a `file://`, x86_64-only, `deadbeef`-hash placeholder, which is why its operator has to run in a direct-bypass mode instead of under the manager.

## Change Class

- Selected class: Class D (core protocol — `src/core/`, facet selector surface)
- New owner-gated state mutator on the blueprints-management facet + a lifted shared helper in `Base`.

## Behavior Contract

- Before: a blueprint's sources are immutable after `createBlueprint`. A bad/placeholder source can never be fixed; the blueprint can't be cold-started by the manager.
- After: `setBlueprintSources(blueprintId, sources)` replaces the stored sources when `msg.sender == blueprint.owner`. Same validation as `createBlueprint`: ≥1 source, each with ≥1 binary, every binary a non-zero `sha256`. Emits `BlueprintSourcesUpdated(blueprintId, sourceCount)`.
- Invariants: non-owner cannot call (reverts `NotBlueprintOwner`); unknown blueprint reverts `BlueprintNotFound`; the published/active `BinaryVersion` set and all other blueprint state are untouched; a re-set fully replaces prior sources (delete-before-write — no stale nested binaries). Sources are intentionally not lockable; ownership is the gate (consistent with `setBlueprintResourceRequirements`).
- The store + source-validation loops are lifted to `Base` as `internal` helpers (`_writeBlueprintSources` / `_validateBlueprintSources`) and shared by `createBlueprint` and `setBlueprintSources` so the two paths cannot drift.

## Risk and Scope

- Blast radius: blueprints-management facet + `Base` helpers + the create-path refactor (behavior-identical). New selector registered on `TangleBlueprintsManagementFacet.selectors()` (12→13 index) and `ITangleBlueprints`.
[...]

PR #163, opened 2026-05-30
fix(blueprints): getBlueprintDefinition reflects live sources, not genesis blob

tangletools:
setBlueprintSources (#162) updates the canonical _blueprintSources, but the blueprint-manager reads getBlueprintDefinition() which returned the stale genesis blob — so repointed sources never reached the manager. Overlay live _blueprintSources onto the decoded definition (single source of truth; no compat shim — testnet/alpha, prior behavior was simply wrong for mutable sources). Manager untouched (it reads the full definition for jobs/schemas too). +1 test.

### DIFF
--- a/src/core/BlueprintsManage.sol
+++ b/src/core/BlueprintsManage.sol
@@ -66,7 +66,7 @@ abstract contract BlueprintsManage is Base {
         return _blueprintMasterRevisions[blueprintId];
     }
 
-    /// @notice Retrieve the original blueprint definition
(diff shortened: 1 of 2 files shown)


drewstone (approved):
## ✅ Auto-approved tangletools PR — `39626208`

This PR was opened by the trusted `tangletools` automation account.

<sub>tangletools · auto-approval · reason: `tangletools_author` · 2026-05-30T21:42:02Z</sub>

PR #169, opened 2026-06-13
feat(deploy): decide mainnet parameter set + wire slash/payment/governance deploy

drewstone:
## What

Decides the Base-mainnet economic + security parameter set from an adversarial multi-expert review (game theory / staking economics / tokenomics / collateral risk / governance / marketplace — 49 agents, every finding red-teamed against the real code bounds), fixes a verified correctness bug, and wires the deploy steps so the decided values actually reach mainnet instead of shipping insecure defaults.

Full rationale + decision table + open-decisions for Drew: **`deploy/MAINNET-PARAMETERS.md`**.

## Verified bug fixed

**Lock-expiry unit bug.** `LockInfo.expiryBlock = block.number + lockDuration` added a *seconds* duration to a *block number* and compared against `block.number`. On Base (~2s blocks) every lock lasted ~2× its label and was chain-block-time dependent. Now `expiryTimestamp` using `block.timestamp` across `DepositManager` + `DelegationManagerLib`; the one block-based test advance switched to `warp`.

## Decided parameters (`deploy/config/base-mainnet.json`)

- Reward multipliers flattened to **1.0×** (field is write-only/unread; prior schedule was risk-*ascending* — a latent landmine a future wiring PR would activate verbatim).
- Anti-sybil operator floor **2 → 1000 TNT**.
- Inflation weight rerouted to the security layer: **staking 5000→5500, developers 1500→1000** (sum stays 10000).
- Gated-asset caps right-sized (EIGEN 1M→250k, USDe 15M→5M, stETH/wstETH 30k→15k as one Lido-family budget).
- New **`governance` / `slashing` / `payments`** blocks consumed by the deploy.

## Deploy wiring (so the decided values deploy)

- `FullDeploy` now calls **`setSlashConfig`** (maxSlashBps 50% / disputeBond 0.02 ETH / resolutionDeadline 21d / maxPending 8) and **`setPaymentSplit`** (keeperBps 50, carved from protocol) during the bootstrap window before role handoff. Without this, mainnet shipped `maxSlashBps=100%`, `disputeBond=0`, `keeperBps=0`.
- New **`script/DeployGovernance.s.sol`** (wraps the audited `GovernanceDeployer`): deploys token+Timelock+Governor from the `governance` block, wires the Governor as proposer/executor/canceller, renounces the bootstrap admin, fails closed on prod against testnet param leakage. `FullDeploy` does not deploy the Governor/Timelock yet `roles.timelock` requires one. Runbook updated with the 0b pre-step.

[...]

PR #170, opened 2026-06-13
ci(foundry): drop unbuildable full test job, keep EIP-170 size gate

drewstone:
## Why

The full `forge build`/`forge test` cannot run on a stock GitHub runner. The contracts require via-IR (it's on in every profile — disabling hits stack-too-deep), and the whole src+test+script graph (~370 files) compiles in one solc unit. That cold via-IR compile:
- OOM-kills the 16GB runner at ~6 min (exit 143), and
- even with the runner disk reclaimed + a 40GB swapfile (~54GB RAM+swap) thrashes the VM unresponsive until GitHub terminates it at ~18 min.

So the cold build never completes, which means an incremental build cache can never seed (chicken-and-egg). A full-suite CI gate needs a larger-memory runner.

## What

- Remove the `Build + Test (fast profile)` job.
- Keep `EIP-170 facet sizes (optimizer)` (`ci_size` profile: src-only, optimizer+via-IR, ~2 min, within RAM) as the sole CI gate — it catches compile breakage and over-limit facets.
- The correctness suite runs locally (`FOUNDRY_PROFILE=fast forge test --no-match-path 'test/FacetSize.t.sol'`, currently 363 passing) and via external audit. Restoring a full gate = point a job at a larger-memory runner + add `forge test`.

Net: `.github/workflows/foundry.yml` only (16 insertions, 80 deletions). The mainnet-params + lock-fix + governance-deploy work already merged via #169.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -1,15 +1,21 @@
-# Foundry CI — test gate for the Solidity protocol.
+# Foundry CI — compile + size gate for the Solidity protocol.
 #
-# Replaces the job removed in #151, which OOM-killed the 16 GB ubuntu runner on
-# `forge build --sizes` (optimizer + via-IR over the full 352-file graph at the
-# ~20-minute mark). This restores the *correctness* gate without that compile:
-#   - Tests run under FOUNDRY_PROFILE=fast (optimizer OFF, sparse_mode), which
-#     fits the 16 GB runner and is the same profile the repo uses locally.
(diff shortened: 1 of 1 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `3c855554`

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-13T23:40:55Z</sub>

PR #186, opened 2026-06-19
fix(staking,payments): token/deposit-unit-consistent amounts for adapter-backed assets

drewstone:
## Summary

Two **adapter unit-mismatch** bugs that surface only for **non-1:1 (rebasing) adapters** (`StandardAssetAdapter` is 1:1, so the existing suites are unaffected — but `RebasingAssetAdapter` is registerable, so these are live, not latent). Both validated end-to-end against the code before fixing.

## HIGH — `LiquidDelegationVault` mints shares off the raw token amount

`deposit()`/`mint()` computed shares from the **raw token `assets`**, but `totalAssets()` is denominated in staking **deposit-units (adapter shares)**, and the delegation pool actually grows by the `credited` deposit-units that `_depositAndDelegate` already computes (and then discarded for the mint). For a rebasing adapter `raw != credited`, so a depositor after a positive rebase minted **more shares than they funded**, diluting earlier holders.

**Fix:** snapshot the pre-deposit rate, then mint against the **credited** deposit-units. `mint()` converts the required deposit-units to the raw token cost via the adapter (`_depositUnitsToAssets`) and mints on `credited`. 1:1 assets are unchanged.

(The finding's headline — "not share-adjusted" — is imprecise; `deposit()` does call `convertToShares`. The real defect is the **unit** of its numerator: raw tokens vs the deposit-unit denomination of `totalAssets()`/`credited`.)

## MEDIUM — `PaymentsEffectiveExposure` feeds deposit-units into a token-unit stat

`getOperatorStakeForAsset` returns **deposit-units** (adapter shares; `_operatorDelegatedAggregate` is incremented by credited units), which `_calculateEffectiveExposures` fed straight into `oracle.toUSD(token, amount)` — `toUSD` expects **token units**. Under a rebasing adapter the USD-normalized exposure (and thus the operator's payout weight) was mis-priced.

**Fix:** convert deposit-units → tokens via `adapter.sharesToAssets` (`_depositUnitsToTokenAmount`) before `toUSD`; 1:1 for native / non-adapter assets; fails open to the input on a missing adapter so a misconfig degrades rather than reverting distribution.

## Tests (`test/audit/batch3/AdapterUnitMismatch.t.sol`)

- **F1** full-stack: real staking diamond + `RebasingAssetAdapter` + `LiquidDelegationVault`; user1 deposits at 1:1, the token rebases +100%, user2 deposits the same nominal amount → mints **~half** the shares (pre-fix: equal → dilution). The `assertLt` fails under the bug.
- **F2** harness: exposes `_calculateEffectiveExposures` with mocked staking/adapter/oracle; only the **token-unit** `toUSD` call is mocked, so the pre-fix deposit-unit path reverts → exposure is asserted to be priced on converted token units.

No regressions: LiquidDelegation, AssetAdapter, LiquidVault, AdaptersMedLow, AdapterChange, RebasingShareScalePoC, Payments, PerAssetExposure, RFQ/Quote distribution, Integration, EndToEndSubscription, MultiAssetDelegation (243 tests).

> Full ~1855-test suite not run in one shot (full via-IR graph exceeds the box's forge memory cap); coverage run per affected subsystem. CI can run the full suite.

### DIFF
--- a/src/core/PaymentsEffectiveExposure.sol
+++ b/src/core/PaymentsEffectiveExposure.sol
@@ -4,6 +4,12 @@ pragma solidity ^0.8.26;
 import { Types } from "../libraries/Types.sol";
 import { IStaking } from "../interfaces/IStaking.sol";
 import { IPriceOracle } from "../oracles/interfaces/IPriceOracle.sol";
+import { IAssetAdapter } from "../staking/adapters/IAssetAdapter.sol";
+
+/// @notice Minimal view of the staking diamond's adapter registry (see `DepositManager`).
+interface IAdapterLookup {
+    function getAssetAdapter(address token) external view returns (address);
+}
 
 /// @title PaymentsEffectiveExposure
 /// @notice Mixin for computing effective exposure (delegation × exposureBps) for payment distribution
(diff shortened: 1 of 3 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `0540f157`

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-19T21:39:32Z</sub>

PR #189, opened 2026-06-24
fix(migration): forbid emergencyWithdraw of the migration token

drewstone:
## What

`TangleMigration.emergencyWithdraw` now reverts (`EmergencyWithdrawTokenForbidden`) when `_token` is the migration token. The migration asset can only ever leave the contract via the permissionless `sweepUnclaimedToTreasury` → immutable `treasury`. Foreign-token and ETH rescue is unchanged.

## Why

`emergencyWithdraw(_token, _amount)` is `onlyOwner` and sends to `owner()` (transferable via `Ownable`). It had **no guard against `_token == address(token)`**, so once `paused` or past the claim deadline the owner could pull the **migration token itself** — i.e. unclaimed user allocations — straight to `owner()`.

That bypassed the entire trustless design: `treasury` is `immutable` and `sweepUnclaimedToTreasury()` is callable by **anyone** after the deadline, specifically so unclaimed funds can only land in a fixed treasury and no privileged actor can redirect them. `emergencyWithdraw` silently reintroduced full owner discretion over those same funds.

In production `owner` will be a multisig Safe, which lowers exploitability to "signers collude / Safe compromised" — but the guard makes the **code match the promise** the contract advertises to users, rather than relying on the signers' good behavior. Cheap, no downside.

This is audit finding #7 (graded medium). The originally-proposed remediation ("route emergencyWithdraw to treasury") was rejected because it would break the legitimate rescue of mis-sent foreign tokens; guarding the migration token specifically closes the hole without that cost.

## Tests

Replaced the prior test that withdrew the migration token to owner (the buggy behavior) with:
- `test_EmergencyWithdraw_RevertsForMigrationToken` — migration token is forbidden even when paused
- `test_EmergencyWithdraw_RescuesForeignToken` — a foreign token is still rescuable to owner
- `test_EmergencyWithdraw_RevertIfNotPausedOrDeadlinePassed` — paused/deadline gate still enforced (via a foreign token)

`forge test --match-path test/TangleMigration.t.sol` → **41 passed, 0 failed**.

### DIFF
--- a/packages/migration-claim/src/TangleMigration.sol
+++ b/packages/migration-claim/src/TangleMigration.sol
@@ -111,6 +111,7 @@ contract TangleMigration is Ownable, ReentrancyGuard {
     error AdminClaimWindowClosed();
     error InvalidAdminClaimDeadline();
     error EmergencyWithdrawNotAllowed();
+    error EmergencyWithdrawTokenForbidden();
     error ClaimDeadlineNotPassed();
     error NoClaimDeadlineSet();
     error MerkleRootLocked();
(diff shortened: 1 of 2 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `66e0b495`

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-24T18:23:08Z</sub>

PR #191, opened 2026-06-24
fix(services): block subscription termination while an earned period is unbilled

drewstone:
## What

Owner-path `terminateService` now reverts `SubscriptionPeriodUnbilled` while a subscription has a fully-elapsed, still-billable, operator-owed period. Anyone can clear it with the permissionless `billSubscription(serviceId)`, after which the owner terminates and reclaims only the unserved tail.

## Why

A subscription owner could terminate inside the keeper's billing-latency window (period elapsed, not yet billed) and `withdrawRemainingEscrow` to reclaim the **entire** escrow — including the period operators had already served. Once status flips to `Terminated`, `_billSubscriptionImpl` rejects the service (`status != Active`), so the earned period becomes permanently unbillable and operators lose those fees. Confirmed real by a red-team exploit PoC.

## Design

- **Gate, don't auto-settle.** In-place settlement during terminate is impossible here: `terminateService` runs in `TangleServicesLifecycleFacet` and billing lives in `TanglePaymentsFacet` — separate facets, separate delegatecall code paths. An internal-call hook compile-conflicts and silently no-ops at runtime. So the lifecycle facet computes eligibility from shared storage and reverts, directing the caller to bill first (permissionless, anyone can).
- **Owner path only.** The gate is in `terminateService`, not `_terminateService`, so `terminateServiceForNonPayment` — which fires precisely when escrow can't cover a period and therefore can *never* be billed — is not deadlocked.
- **Conditions mirror `PaymentsBilling._billSubscriptionImpl`** (subscription, within TTL, full interval elapsed since `lastPaymentAt`) and must be kept in lockstep. A non-empty operator set is a conservative "fees are owed" proxy; an all-inactive set just costs one harmless cursor-advancing bill before terminate proceeds.

## Verification

- New regression `test_Regression_TerminateBlockedUntilEarnedPeriodSettled`: terminate reverts while period unbilled → permissionless bill settles ~1 period to operators → owner then terminates and is refunded only the unserved tail. **Passes.**
- Broader subscription + lifecycle suites: **132/132 pass** (no existing terminate/billing flow, including non-payment termination, regressed).

## Lineage

Third and final fix from the red-team pass. Siblings: `emergencyWithdraw` token guard (#189, merged) and vesting delegatee front-run (#190). The "HIGH" cross-operator cost-basis claim was debunked (its PoC failed its own drain assertion).

PR #190, opened 2026-06-24
fix(migration): bind vesting delegatee to beneficiary to stop front-run capture

drewstone:
## What

`TNTVestingFactory.getOrCreateVesting` no longer accepts a caller-supplied `delegatee`; it forces `delegatee == beneficiary`. Deploy scripts and `FullDeploy` updated to the 3-arg signature.

## Why

The `delegatee` was **not part of the CREATE2 salt** (`keccak256(token, beneficiary, startTimestamp)`). Since `getOrCreateVesting` is permissionless, a front-runner could pre-create a claimant's vesting clone at the deterministically-predicted address with an **attacker delegatee**. The migration's `_executeClaim` then found the clone already deployed, skipped initialization, and merely funded it — redirecting the beneficiary's **vested voting power** to the attacker (for an `ERC20Votes` token such as the production `TangleToken`).

Tokens were never at risk — `beneficiary` is salt-bound, so the attacker can't steal the vested tokens. Only governance voting power was redirected, and only until the beneficiary noticed and re-delegated. Real medium-severity governance griefing, especially at TGE scale across many claimants.

Binding the delegatee to `beneficiary` removes the attacker-controlled degree of freedom: any pre-created clone is now byte-identical to the one the migration would create, so a front-run is a harmless no-op. The beneficiary may still re-delegate afterward via the vesting contract.

## How found

Red-team exploit PoC. The PoC is retained as a passing regression test: the attacker pre-creates the clone, but after the victim's claim the attacker holds **0 votes** and the beneficiary holds **all** of the vested voting power.

## Verification

- `packages/migration-claim`: **42/42 tests pass** (`forge test`), including the new regression.
- Main repo compiles clean with the `FullDeploy` 3-arg fix.

## Scope note

The two `getOrCreateVesting` deploy-script call sites (treasury / foundation vesting) already passed `delegatee == recipient`, so dropping the argument is behavior-preserving.

PR #192, opened 2026-06-25
test(security): cost-basis solvency regression + live front-run fork proof

drewstone:
Closes the two open questions left after the red-team pass. Both are passing tests.

## 1. The "HIGH" is provably not exploitable (`Regression_Delegation_CostBasisSolvency`)

A red-team agent rated `DelegationManagerLib.sol:616-694` a HIGH: a delegator holding two same-asset positions to operators with divergent exchange rates supposedly under-realizes a slash loss and drains other delegators. The original exploit PoC **failed its own over-withdraw assertion** — `dep.amount` is internally overstated after the slash, but the withdrawable amount stayed correctly bounded.

This regression exercises the full lifecycle the original stopped short of: the attacker fully exits **both** positions and withdraws everything, then the honest co-delegator withdraws. Two decisive invariants both hold:
- the attacker cannot extract more than their true post-slash entitlement (opX 10 + opY 5 = **15 ETH**) — non-vacuity guarded (they really do recover ~15, so the bound is a real test;
- the honest co-delegator is still made whole (**30 ETH**) — the pool is solvent.

Conclusion: the internal `dep.amount` overstatement is fully offset and **not monetizable**. Not a vulnerability.

## 2. The front-run is live in production (`Fork_MigrationClaim_VestingFrontRun_BaseSepolia`)

Proves the vesting delegatee front-run (fixed in #190) is real against the **actual Base Sepolia deployment**, not just the local harness. The test forks the chain, calls the deployed `TNTVestingFactory` (`0x45C1…`), and shows a victim's vesting clone delegating its voting power to an attacker on the real ERC20Votes `TangleToken` (`0xB546…`). Confirms #190 is needed in prod.

Forking degrades gracefully (`vm.skip`) when the RPC is unreachable or the deployment is absent, so it never red-fails a sandboxed/offline CI runner. Verified locally against `https://sepolia.base.org`: **PASS, 0 skipped**.

## Context

Final cleanup from the red-team pass. Fixes: `emergencyWithdraw` token guard (#189, merged), vesting front-run (#190), subscription terminate gate (#191).

### DIFF
--- a/test/exploit/Regression_Delegation_CostBasisSolvency.t.sol
+++ b/test/exploit/Regression_Delegation_CostBasisSolvency.t.sol
@@ -1,0 +1,171 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.26;
+
+import { Test } from "forge-std/Test.sol";
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+
+import { IMultiAssetDelegation } from "../../src/interfaces/IMultiAssetDelegation.sol";
+import { MultiAssetDelegation } from "../../src/staking/MultiAssetDelegation.sol";
+import { Types } from "../../src/libraries/Types.sol";
+
+import { StakingOperatorsFacet } from "../../src/facets/staking/StakingOperatorsFacet.sol";
+import { StakingDepositsFacet } from "../../src/facets/staking/StakingDepositsFacet.sol";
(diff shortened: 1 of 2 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `5a1e9808`

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-25T08:01:12Z</sub>

PR #193, opened 2026-07-02
feat(blueprints): hash+event-source the blueprint definition — createBlueprint gas -64%

drewstone:
## Problem

`createBlueprint` stored the full blueprint definition **two-to-three times** in 20k-gas-per-slot storage — a redundant `abi.encode(def)` blob, the same data decomposed into typed fields, and a write-only cross-contract record in `MasterBlueprintServiceManager` — plus **16 human-readable display strings** (job names/descriptions/URIs and 7 metadata prose fields) that **no on-chain rule ever reads**.

On mainnet this is waste. On chains that meter storage far above mainnet (Tempo ~6–11× SSTORE, 30M gas/tx cap), it made `createBlueprint` **exceed the block cap and OOG (>28M) — blueprint registration was impossible.**

## Change — store a hash, emit the data, keep only what on-chain logic reads

A datum belongs in storage only if a trustless invariant reads it on-chain. Everything else is display/discovery and belongs in an event, anchored by a keccak digest.

- **Definition blob → digest.** `_blueprintDefinitionBlobs` (full `abi.encode` blob) is gone; `_blueprintDefinitionHash` (`bytes32`) anchors the definition. The full bytes are already emitted in `BlueprintDefinitionRecorded`; new `blueprintDefinitionHash(id)` view lets any consumer verify the event copy.
- **MBSM record → digest.** `MasterBlueprintServiceManager` stores `definitionHash`, not the blob (it already emits the event).
- **Display data → event.** Job `name`/`description`/`metadataUri` and metadata `description/author/category/codeRepository/logo/website/license` are no longer stored. **Functional fields stay on-chain**: job param/result schemas (`validatePayload`), sources (operator binary resolution), config, `metadata.name` + `profilingData` (manager reads them), `metadataUri` + `metadataHash`.
- **Job storage** now carries the full `JobDefinition` shape (schemas populated, display strings empty) so `getBlueprintDefinition` round-trips job count/order/schemas exactly.
- **Manager hook budget 500k → 2M.** A real BSM setup hook does a few SSTOREs; under high-metering chains that exceeds 500k and reverts `createBlueprint`. 2M clears it while staying a hard reentrancy/DoS bound (new `ManagerHookGasBudget` test pins both directions).
- **Deploy gap fixed.** `FullDeploy` now grants the Tangle `RECORDER_ROLE` on the metrics contract (the pointer was set but the role never granted, so `recordBlueprintCreated` reverted-but-caught on every create).
- **Greenfield cleanup.** Removed two dead reserved slots (`_pendingNativeRewards`, `_deprecatedTntStakerFeeBps`); folded the definition hash into a clean layout.
- **Build.** `dynamic_test_linking` on the default profile (fixes the full-graph via-IR OOM / stack-too-deep no single contract reproduces); off in `profile.fast` since linking alters `vm.prank`+`new` CREATE semantics.

## Gas (measured on-chain, Tempo Moderato, full 5-job sandbox blueprint)

| Stage | createBlueprint gas | outcome |
|---|---|---|
| Original (blob stored) | **>28M** | OOG — registration impossible |
| Blob → hash | 12.65M | fits the 30M cap |
| Display → event (this PR) | **~10M** (9.69–10.19M) | −64% vs original |

Mainnet-equivalent createBlueprint ≈ 0.55M. `blueprintCount=3` on the greenfield redeploy, all txs status 1.

## SDK

The full definition is already in `BlueprintDefinitionRecorded`. The Rust blueprint SDK reads the dropped display fields from that event, verified against `blueprintDefinitionHash` (companion change in `tangle-network/blueprint`). The **operator manager path is unchanged** — it reads only fields that stay on-chain (`metadata.name`, `profilingData`, `sources`, schemas). Only the CLI's *display* reads move.

## Tests

- 96 blueprint tests incl. exact job round-trip, event-anchor integrity (payload keccak == on-chain hash), metadata-split, and the manager-hook budget bounds.
- Full suite green (1879+); size gate under EIP-170 (`ci_size --sizes`).
- Greenfield Tempo redeploy verified end-to-end (FullDeploy clean at 105% under the 30M cap).


tangletools (dismissed):
## ✅ Auto-approved drewstone PR — `fb6a47ba`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T05:17:06Z</sub>

tangletools (commented):
## 🟠 Value Audit — better-approach-exists

| | |
|---|---|
| **Verdict** | better-approach-exists |
| **Concerns** | 3 (1 strong-concern, 2 weak-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.1s |
| **Interrogation** | 276.9s (2 bridge agents) |
| **Total** | 277.0s |

### 💰 Value — better-approach-exists

Replaces stored blueprint definition blobs with a 32-byte hash plus an event-sourced copy, drops unread display strings from on-chain storage, raises the manager-hook gas budget, and moves router-only admin/views out of Base; the core idea is sound, but deleting deprecated storage slots breaks UUPS 

- **What it does:** In createBlueprint, it no longer SSTOREs the full abi.encode(def) blob or a copy in the master manager; it stores only keccak256(encodedDefinition) (src/core/BlueprintsCreate.sol:56-61), emits the full bytes in BlueprintDefinitionRecorded, and stores only the functional decomposed fields (config, schemas, sources, memberships, metadata.name/profilingData/metadataUri/metadataHash). Job definitions 
- **Goals it achieves:** Slash createBlueprint gas (~64%) so blueprint registration fits under high-metering chain block caps; anchor off-chain discovery/indexing to a verifiable hash; keep only data that on-chain rules actually read; give real BSM setup hooks enough gas on chains like Tempo; and reduce facet bytecode by moving router-only functions out of Base.
- **Assessment:** The hash-and-event approach is coherent and follows the codebase's existing event-sourcing pattern. Job/schema round-trips are preserved, the MBSM digest matches the event payload, and the gas-budget change is justified. However, deleting _pendingNativeRewards and _deprecatedTntStakerFeeBps while leaving __gap at 28 rewrites the UUPS storage layout for every downstream variable, which is unsafe fo
- **Better / existing approach:** Keep the deprecated slots as reserved variables (or rename them to _reservedX / insert explicit gap placeholders at the exact deleted positions) instead of removing them, preserving the storage layout for all variables declared after the rewards block. This is the standard pattern the repo already uses for upgrade safety (see the explicit __gap and append-only storage comments in TangleStorage.sol
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

### 🎯 Usefulness — sound-with-nits

Event-sources the blueprint definition (store a keccak digest, emit the bytes) and drops on-chain-write-only display fields to unblock createBlueprint on high-metering chains like Tempo — coherent, reachable, and the indexer already consumes the event so nothing breaks.

- **Integration:** Fully wired and reachable. createBlueprint (BlueprintsCreate.sol:26) is the hot path and every caller now hits the new hash-store. The new blueprintDefinitionHash view is both defined (BlueprintsManage.sol:79) and registered as a facet selector (TangleBlueprintsManagementFacet.sol:32) plus exposed via ITangleBlueprints.sol. Critically, the Envio indexer already event-sources the full definition — 
- **Fit with existing patterns:** Hits the codebase's grain on both counts. (1) Store-hash-emit-bytes is the canonical Solidity gas pattern for data no on-chain invariant reads; grep confirms zero on-chain readers of the dropped display fields (job name/description/metadataUri; metadata author/category/codeRepository/logo/website/license/description), while functional fields (paramsSchema/resultSchema read by SchemaLib.validatePay
- **Real-world viability:** Holds up past the happy path. getBlueprintDefinition round-trips job count/order/schemas exactly (positional job index stays load-bearing for submissions); only display strings return empty, which is documented and SDK-event-sourced. The hook-budget raise 500k→2M is bounded by a hard cap and tested against both a realistic heavy-setup BSM (~1.2M gas) and a hostile unbounded-gas BSM that the cap mu
- **Bridge attempts:** 1

### 🎯 Usefulness Audit

🟡 **getBlueprintDefinition now hardcodes hasConfig=true for config-less blueprints** `[ergonomics]` ``
> BlueprintsManage.sol:100 sets definition.hasConfig = true unconditionally. A blueprint created with hasConfig=false gets its config normalized to Fixed/PayOnce (BlueprintsCreate.sol:157-160), so the view now reports hasConfig=true even though the caller never supplied one — a semantic drift vs. the creation-time value. It is documented in the view's docstring and is view-only, so it does not affect on-chain logic; but any off-chain consumer that branches on hasConfig to decide whether the creato

### 💰 Value Audit

🔴 **Deleting deprecated storage slots breaks UUPS upgrade layout** `[better-architecture]` ``
> TangleStorage.sol removes _pendingNativeRewards and _deprecatedTntStakerFeeBps (git diff HEAD~1 -- src/TangleStorage.sol) but leaves __gap as uint256[28]. Because this is a UUPS upgradeable system (Base.sol:28, Tangle.sol), removing a mapping and a packed uint16 shifts every subsequent storage variable (Slashing, RFQ, Metrics, Incentives, etc.) into the slots previously occupied by those variables. The correct pattern is to keep deprecated slots as reserved variables or explicit gap placeholders

🟡 **PR bundles unrelated Base split and ServiceFeeDistributor refactor with the blueprint storage change** `[proportion]` ``
> The diff also introduces TangleCoreApi.sol (moving ~440 lines of Base admin/views) and refactors ServiceFeeDistributor.sol distribution helpers. Both are reasonable in isolation, but they expand the review surface beyond the stated 'hash+event-source blueprint definition' goal. If kept, they should be clearly tied to the same deploy branch; otherwise they risk being backported without independent review.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T052551Z</sub>

drewstone:
Thanks — addressing the audit:

**🔴 Dead-slot deletion / UUPS layout:** intentional. This is a **greenfield** protocol — every deployment is a test that we redeploy from scratch (`FullDeploy`), never an in-place proxy upgrade, so the layout shift is safe by construction and carrying reserved deprecation slots is explicitly what we're removing. Once we cut a mainnet release the layout freezes and append-only discipline resumes; pre-launch we're finalizing it clean.

**🟡 hasConfig=true in the view:** correct catch — the creation-time `hasConfig` is lost to normalization. Following up with a packed `hasConfig` bool on the `Blueprint` struct (same slot as `active`, ~free) so `getBlueprintDefinition` reports the real value.

**🟡 Bundled Base split + ServiceFeeDistributor:** same deploy branch on purpose — the `TangleCoreApi` split is the EIP-170 headroom that let the Tempo facets fit, and both landed while getting createBlueprint under the 30M cap. Tied to this deploy, not backport-able independently.

tangletools (approved):
## ✅ Auto-approved drewstone PR — `0ea2efa1`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T05:52:14Z</sub>

tangletools (commented):
## ⚪ Value Audit — audit-incomplete

| | |
|---|---|
| **Verdict** | audit-incomplete |
| **Concerns** | 0 (none) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 36.5s (2 bridge agents) |
| **Total** | 36.8s |

### 💰 Value — error

value agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/zai-coding-plan/glm-5.2: Bridge agent streaming failed: [Errno 111] Connection refused; opencode/deepseek/deepseek-v4-pro: Bridge agent streaming failed: [Errno 111] Connection refused

### 🎯 Usefulness — error

usefulness agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/kimi-for-coding/k2p7: Bridge agent streaming failed: [Errno 111] Connection refused; opencode/deepseek/deepseek-v4-pro: Bridge agent streaming failed: [Errno 111] Connection refused

No PR concerns were produced because the value/usefulness agent pass did not complete. Treat this audit as incomplete, not as approval.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T061622Z</sub>

tangletools:
## ❌ Needs Work — `0ea2efa1`

**Review health 100/100** · **Reviewer score 0/100** · **Confidence 95/100** · 23 findings (1 critical, 2 high, 4 medium, 16 low)

| | opencode-kimi | glm | aggregate |
|---|---|---|---|
| **Readiness** | 0 | 20 | **0** |
| **Correctness** | 0 | 20 | **0** |
| **Security** | 0 | 20 | **0** |
| **Testing** | 0 | 20 | **0** |
| **Architecture** | 0 | 20 | **0** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 8/8 planned shots over 43 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 43 changed files. Global verifier still owns final merge decision.

### Blocking

🟣 **CRITICAL** Inherited storage layout break in upgradeable Tangle proxy — [`src/core/Base.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/Base.sol)
> Base inherits TangleStorage, whose layout changed incompatibly in this PR: (1) _blueprintHasConfig(bool) was inserted between _blueprintConfigs and _blueprintMetadataUri, shifting every subsequent mapping slot; (2) _pendingNativeRewards and _deprecatedTntStakerFeeBps were removed, shifting all later slots; (3) _blueprintDefinitionBlobs(bytes) and _blueprintJobSchemas(StoredJobSchema[]) were replaced with _blueprintDefinitionHash(bytes32) and _blueprintJobs(JobDefinition[]) at the same positions. For a UUPS upgrade this corrupts existing blueprint metadata, config, sources, job schemas, and every state variable stored after the shifted slots. This is concrete from the old/new TangleStorage diff and the shot files' use of these variables (e.g., BlueprintsCreate.sol:51-62). Fix: append new va

🔴 **HIGH** Non-backward-compatible storage layout lacks inline warning — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol)
> Insertion of `_blueprintHasConfig` shifts every subsequent mapping root by +1 slot: `_blueprintMetadataUri` 8→9, `_blueprintMetadataHash` 9→10, `_blueprintSources` 11→12, `_blueprintDefinitionBlobs`/`_blueprintDefinitionHash` 14→15, `_blueprintJobSchemas`/`_blueprintJobs` 22→23, `_serviceRequests` 23→24, and 25 other mappings through `_pendingRewards` 44→45 (verified with `forge inspect Tangle storageLayout` base vs head). Removed `_pendingNativeRewards` and `_deprecatedTntStakerFeeBps` also change packing. The commit message calls this 'greenfield, no back-compat', but the contract natspec/commentary gives no warning, creating an accidental-upgrade footgun for the existing base-sepolia UUPS proxy. Add a prominent top-of-file or section comment: this layout is greenfield-only and must not 

🔴 **HIGH** Existing blueprint blobs/job arrays corrupted by same-slot type replacement — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsCreate.sol#L62)
> At [line 62](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsCreate.sol#L62) the code stores keccak256(encodedDefinition) into _blueprintDefinitionHash, occupying the same storage variable position formerly used by _blueprintDefinitionBlobs(bytes). For any pre-existing blueprint, that slot holds a bytes length word (or packed short bytes), not a keccak256 digest. Similarly, _blueprintJobs at [line 81-97](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsCreate.sol#L81-L97) replaces _blueprintJobSchemas(StoredJobSchema[]) with a different struct layout at the same position. After an upgrade, getBlueprintDefinition would return garbage metadata/config/job data and job submiss

### Other

🟠 **MEDIUM** Contract docs claim upgrade-compatibility but this PR makes a breaking greenfield layout reset — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol#L14)
> The contract-level @dev ([line 14](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol#L14)) still reads 'Inherit this contract to maintain storage compatibility across upgrades' and [line 15](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol#L15) 'Storage slots are explicitly managed to prevent collisions'. This PR inserts _blueprintHasConfig at slot 8 (shifting every subsequent field by +1), removes _pendingNativeRewards (shifting the tail by -1), removes the packed _deprecatedTntStakerFeeBps (changing _tntPaymentDiscountBps's byte offset within its slot), and swaps two value types in place. The commit fb6a47b explicitly declares 'greenfield, no back-compat', so the cha

🟠 **MEDIUM** StorageLayoutSnapshotTest pins only slots 0-7, missing the entire reshuffled region — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol)
> test/security/StorageLayoutSnapshotTest.t.sol:44-51 pins _staking(0) through _blueprintConfigs(7). The new _blueprintHasConfig sits at slot 8, immediately AFTER the last pinned field, so the test passes green despite ~80 subsequent slots shifting (+1 from the insertion, -1 from _pendingNativeRewards removal). The test's own header says it exists to force authors to 'move the new field to the end ... consume a __gap slot, or explicitly bless the new layout by updating this snapshot' — none of which this PR did, yet the test cannot detect it. Post-greenfield this snapshot test is the ONLY CI guard against accidental future mid-layout insertions, and it is structurally blind to any change past slot 7. Fix: pin at least one tail field (e.g. __gap start slot, _slashCommitmentSnapshots, _pending

🟠 **MEDIUM** MANAGER_HOOK_GAS_LIMIT quadrupled (500k→2M), significantly widening BSM DoS surface on all chains — [`src/core/Base.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/Base.sol)
> The constant was raised from 500_000 to 2_000_000. The comment justifies this for Tempo chain where SSTORE is ~11x mainnet, but the constant applies to ALL chains (mainnet, Base, Arbitrum). On cheaper-gas chains, a malicious or buggy BSM can now burn 2M gas per hook invocation — 4x the previous ceiling. This affects every _callManager / _tryCallManager / _tryStaticcallManager call path (PaymentsDistribution.sol:463, PaymentsBilling.sol:469, JobsSubmission.sol:234, etc.), including batch paths like submitResults where the hook fires per-result. nonReentrant guards on entry points prevent reentrancy, and the ManagerHookGasBudget.t.sol test proves the hostile BSM is still capped, but the increased per-hook cost floor could make legitimate high-fanout transactions (e.g., batch job results with

🟠 **MEDIUM** Manager hook gas budget quadrupled, expanding DoS/reentrancy surface — [`src/core/Base.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/Base.sol)
> MANAGER_HOOK_GAS_LIMIT was raised from 500_000 to 2_000_000. The comment justifies this for chains with higher SSTORE pricing, but the constant caps every state-changing and static manager hook in _callManager, _tryCallManager, and _tryStaticcallManager. A malicious or buggy BSM can now consume up to 2M gas per call, and multiple hooks can fire in one transaction (e.g., job result + aggregation finalization). While still capped, the 4x increase meaningfully expands the worst-case gas griefing and reentrancy window. The new test (ManagerHookGasBudget.t.sol) only verifies the cap is enforced, not that legitimate protocol paths still have sufficient gas after a 2M hook. Fix: bound the total gas forwarded across all hooks in a single transaction, or keep a lower default and make the elevated T

🟡 **LOW** Breaking ABI selector changes for 3 quote functions — off-chain consumers must rebind — [`bindings/src/bindings/i_tangle_services.rs`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/bindings/src/bindings/i_tangle_services.rs)
> The function selectors changed due to signature modifications: createServiceFromQuotes 0x458a69a0->0x8ccea21e, extendServiceFromQuotes 0xebb7d84b->0x5d91ea20 (i_tangle_services.rs), submitJobFromQuote 0x52ada2be->0x72a1f7be (i_tangle_jobs.rs). Any off-chain service, indexer, or SDK that hardcodes the old 4-byte selectors or the old struct layouts (QuoteDetails without operation/serviceId, JobQuoteDetails without inputsHash) will silently fail or mis-decode. The bindings correctly track the new selectors — this is inherent to the Solidity ABI break, not a binding bug. Flagging so the global verifier ensures a coordinated release (bindings bump + downstream consumer updates) since tnt-core publishes these bindings as a crate.

🟡 **LOW** New i_master_blueprint_service_manager module missing crate-root re-export in lib.rs — [`bindings/src/bindings/mod.rs`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/bindings/src/bindings/mod.rs#L6-L22)
> mod.rs adds `pub mod i_master_blueprint_service_manager;` but lib.rs (outside this shot's files) does not add the companion `pub use bindings::i_master_blueprint_service_manager::IMasterBlueprintServiceManager;`. Every other module in mod.rs has a matching re-export at the crate root (lib.rs [lines 6-22](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/bindings/src/bindings/mod.rs#L6-L22)). The new type is reachable via the full path bindings::bindings::i_master_blueprint_service_manager:: but is inconsistent with the established pattern. Impact: downstream consumers who rely on the crate-root re-exports (e.g., `use bindings::ITangle;`) will not find `IMasterBlueprintServiceManager` at the same level. Fix: add `pub use bindings::i_master_blueprint_se

🟡 **LOW** Gap provenance comment regressed; audit-trail history lost — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol)
> Old comment: 'Reserved storage slots for future upgrades. Standard gap size is 50. Slots already consumed: 10 (initial) + 5 (binary-versions block) + 5 (supply-chain hardening block above) + 1 (slash commitment snapshot). Shrunk 34 -> 29 -> 28 as those blocks were appended.' New comment: 'Reserved storage slots for future appends.' The arithmetic is correct (net 0 slot delta this PR), but the consumption history that docs/audits/REDTEAM-STORAGE-UPGRADE-2026-05-16.md relied on for its clean-checks section is gone. Re-adding a one-line consumption ledger ('__gap=28; consumed by: initial(10)+binVersions(5)+supplyChain(5)+slashSnapshot(1)+blueprintHasConfig-offset...') preserves the auditability that the prior red-team storage audit depended on.

🟡 **LOW** No deploy-time guard against upgrading the live base-sepolia proxy to this greenfield impl — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol)
> deployments/base-sepolia/latest.json records a live Tangle UUPS proxy at 0x8299d60f373f3a4a8c4878e335cb9d840e6e3730 initialized under the pre-reset layout. Nothing in TangleStorage, Base.initialize, or script/FullDeploy.s.sol prevents an UPGRADER_ROLE holder from calling upgradeToAndCall() on that proxy with this implementation. Doing so would silently misread every field from slot 8 onward (e.g. _blueprintMetadataUri's mapping root shifts, _pendingRewards tail shifts, _tntPaymentDiscountBps reads the old _deprecatedTntStakerFeeBps byte offset). This is an operational risk, not a code bug, since the commit declares greenfield intent — but a cheap defense (e.g. an initialize()-time version constant assert, or a deploy-script check that aborts if the target proxy already has non-zero storage

🟡 **LOW** Storage layout is NOT upgrade-safe — variable insertion + type changes would corrupt existing state — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/TangleStorage.sol)
> _blueprintHasConfig (mapping(uint64=>bool)) is inserted between _blueprintConfigs and _blueprintMetadataUri, shifting all subsequent variables by one slot. Combined with type changes (_blueprintDefinitionBlobs bytes→_blueprintDefinitionHash bytes32 at same slot, _blueprintJobSchemas StoredJobSchema[]→_blueprintJobs JobDefinition[] at same slot with different struct sizes: 2 vs 5 slots/element), and removals (_pendingNativeRewards, _deprecatedTntStakerFeeBps), upgrading any existing deployment would corrupt all blueprint/reward/protocol state. This is acceptable for a fresh deployment but must never be used as an upgrade path. The __gap stays at 28 (net -1 slot consumed), confirming this was designed as a breaking storage change. Ensure deployment scripts reflect this.

🟡 **LOW** Metadata field regression — 7 of 9 BlueprintMetadata fields dropped from on-chain storage — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsCreate.sol)
> _storeBlueprintMetadata now stores only BlueprintMetadata.name and .profilingData, zeroing the other 7 fields (description, author, category, codeRepository, logo, website, license). The blueprintMetadata() view (BlueprintsManage.sol:38-46) and getBlueprintDefinition() return the full BlueprintMetadata struct with these fields as empty strings. Any off-chain consumer (explorer, SDK, dashboard) that previously read description/author/category/etc. from the contract will silently get empty strings. The full metadata is available in the BlueprintDefinitionRecorded event. This is an intentional gas optimization, but it is a breaking API change for any consumer reading metadata on-chain rather than from events.

🟡 **LOW** blueprintDefinitionHash cannot verify the on-chain reconstructed definition — only event-sourced copies — [`src/core/BlueprintsManage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsManage.sol)
> _blueprintDefinitionHash stores keccak256(abi.encode(FULL def)) at creation time. But getBlueprintDefinition returns a PARTIAL def (metadata has 7/9 fields empty, jobs have 3/5 fields empty). Therefore keccak256(abi.encode(getBlueprintDefinition(id))) will NOT equal blueprintDefinitionHash(id). The test_DefinitionHashAnchorsEventSourcedCopy test verifies the hash against the event payload and the MBSM record, but not against the on-chain view. This is by design (the hash anchors the event, not the reconstruction), but the NatSpec should explicitly state this so no consumer attempts to verify the on-chain view against the hash and gets a false mismatch.

🟡 **LOW** getBlueprintDefinition NatSpec 'Jobs round-trip exactly' is misleading — display fields silently emptied — [`src/core/BlueprintsManage.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/core/BlueprintsManage.sol)
> The docstring states 'Jobs round-trip exactly.' But _storeBlueprintSchemas (BlueprintsCreate.sol:82-96) stores JobDefinition with empty name/description/metadataUri. getBlueprintDefinition reconstructs from these fields, so the returned jobs have empty display strings. Only count, order, paramsSchema, and resultSchema survive. The test_JobsRoundTripExactlyThroughDefinitionView test documents this explicitly (asserting name/description/metadataUri are empty), but the NatSpec claim 'round-trip exactly' overpromises — a consumer reading only the docstring would expect all JobDefinition fields to survive. Recommend rewording to 'Job count, index order, and schemas round-trip exactly; display strings are dropped on-chain (carried by the BlueprintDefinitionRecorded event).'

🟡 **LOW** Interface omits sibling master-manager events — [`src/interfaces/IMasterBlueprintServiceManager.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/interfaces/IMasterBlueprintServiceManager.sol)
> This PR promotes BlueprintDefinitionRecorded into the interface but leaves BinaryVersionRecorded and OperatorBinaryAckRecorded declared only in the concrete contract (src/MasterBlueprintServiceManager.sol:34,39). The three events are peers (all 'authoritative indexer events' off the same hook functions onBinaryVersionPublished/onOperatorBinaryAcked/onBlueprintCreated), so consumers reasoning from IMasterBlueprintServiceManager see an incomplete event surface and must also read the impl. Impact: documentation/IDE/discoverability only — no behavioral risk, since events are not ABI-enforced and the indexer config lists all three directly. Fix: add the two missing event declarations to the interface for symmetry, or document why they are intentionally impl-only.

🟡 **LOW** Compile gate not run in this sandbox (environment limitation) — [`src/libraries/SchemaLib.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/libraries/SchemaLib.sol)
> The Foundry binary crashes in this worktree (forge-cap shim cannot locate real forge; ~/.foundry/bin/forge v1.7.1 panics on both build and soldeer update). Dependencies (dependencies/, lib/) are absent. Full compile verification was not possible here. Mitigation: static cross-file verification confirms type consistency at every call site (4 production callers + test harness), the removed struct has zero references, and field types match. CI on the PR should be the authoritative compile gate.

🟡 **LOW** Minor API smell: blueprintId/jobIndex redundant with passed job (pre-existing) — [`src/libraries/SchemaLib.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/libraries/SchemaLib.sol)
> validateJobParams/validateJobResult now receive the whole Types.JobDefinition storage job but still take separate blueprintId and jobIndex params. Since callers fetch the job via _jobDefinition(blueprintId, jobIndex), these two params duplicate information already embedded in how the job was located. They are still needed as context for Errors.SchemaValidationFailed reporting inside _validate (refId/auxId), so this is intentional and pre-existing design — not a regression. No fix required for this PR; noted only for future API simplification (e.g., storing blueprintId/jobIndex inside JobDefinition).

🟡 **LOW** Stale reward debt not zeroed when position score reaches 0 (mode 1) — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L893)
> _settlePositionDebt returns early when userScore == 0 ([line 893](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L893)), so _debtAll is NOT zeroed. The old _syncDebtsToCurrentAcc unconditionally wrote _debtAll = (userScore * acc) / PRECISION, which zeroed it when score was 0. This is PROVABLY HARMLESS: (a) _pendingPosition also returns 0 when userScore == 0, so no incorrect claim can occur; (b) on re-delegation, onDelegationChanged calls _syncDebtsToCurrentAcc ([line 322](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L322)) with the new non-zero score BEFORE any distribution can accrue, overwriting the stale debt; (c) _pruneDelegat

🟡 **LOW** Unused durationSeconds parameter in _distributeChunk after delegation to _distributeImmediate — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L1216)
> After the refactor, _distributeChunk just delegates to _distributeImmediate, making its uint256 /* durationSeconds */ parameter dead. Both callers (_dripOperatorStreams [line 1216](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L1216), drip [line 1247](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/src/rewards/ServiceFeeDistributor.sol#L1247)) still pass it from the streaming manager's return values. Cosmetic cleanup opportunity: drop the param and adjust callers, or inline the _distributeImmediate call at the call sites. Not a correctness issue.

🟡 **LOW** Gas-bound assertion is loose (6M vs 2M cap) and Tempo scenario not exercised — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/test/blueprints/ManagerHookGasBudget.t.sol)
> `assertLt(gasBefore - gasleft(), 6_000_000)` uses a 3x margin over the 2M cap. This is intentionally loose and documented, and correctly catches the 'cap removed entirely' regression. Separately, the StorageHeavySetupBSM comment references Tempo's 6-11x storage pricing as motivation for the raised cap, but the test only validates mainnet gas metering (Foundry default). On Tempo, 54 SSTOREs at 11x would cost ~12M and exceed the 2M cap. This is a test-scope limitation (chain-specific gas pricing is outside EVM contract control), not a code defect — noting it so the Tempo headroom claim isn't mistaken for a tested invariant.

🟡 **LOW** GasExhaustingBSM loop terminates on uint256 wrap, not truly unbounded — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/test/blueprints/ManagerHookGasBudget.t.sol)
> The hostile hook uses `for (uint256 i = 1; i != 0; ++i)` which is practically infinite (2^256-1 iterations) and will always OOG within the 2M cap before wrapping — so the test's intent (verify the cap stops unbounded work) is fully met. The loop condition `i != 0` means it would theoretically halt at the wrap point, but this is astronomically unreachable and irrelevant to test correctness. Pure documentation clarity nit; no functional impact.

🟡 **LOW** Compiler warning: test function can be view — [`test/blueprints/MasterBlueprintServiceManager.t.sol`](https://github.com/tangle-network/tnt-core/blob/0ea2efa13471014ceeab9dce545e10b4d0ef6e6f/test/blueprints/MasterBlueprintServiceManager.t.sol)
> Solc emits Warning 2018: `Function state mutability can be restricted to view` for `test_Constructor_GrantsRoles`. The function only reads roles, so declaring it `public view` removes the warning and matches the read-only nature of the assertion.

---
<sub>tangletools · 2026-07-02T08:00:18Z · [trace](https://gist.github.com/tangletools/fd4335584604c00f3d9e8604b208df5c)</sub>

tangletools (dismissed):
## ❌ 3 Blocking Findings — `0ea2efa1`

> Full multi-shot audit completed 8/8 planned shots over 43 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 43 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/fd4335584604c00f3d9e8604b208df5c)

Summary comment for this run: [full summary](https://github.com/tangle-network/tnt-core/pull/193#issuecomment-4863513251)

---
<sub>tangletools · 2026-07-02T08:00:18Z · immutable [trace](https://gist.github.com/tangletools/fd4335584604c00f3d9e8604b208df5c)</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `3f2b9f05`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T10:44:23Z</sub>

tangletools (commented):
## ⚪ Value Audit — audit-incomplete

| | |
|---|---|
| **Verdict** | audit-incomplete |
| **Concerns** | 0 (none) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 906.5s (2 bridge agents) |
| **Total** | 906.8s |

### 💰 Value — error

value agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

### 🎯 Usefulness — error

usefulness agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

No PR concerns were produced because the value/usefulness agent pass did not complete. Treat this audit as incomplete, not as approval.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T110750Z</sub>

drewstone:
### Audit round addressed (`3f2b9f0`)

**Reconciled (greenfield — not code-changed):** the CRITICAL/HIGH storage-layout findings. This is a pre-mainnet greenfield protocol — every deploy is redeployed from scratch via `FullDeploy`, never an in-place UUPS upgrade — so the layout reset is intended. But I addressed the reviewers' *legit sub-points*:
- **Undocumented layout footgun** → added a prominent GREENFIELD/PRE-MAINNET warning at the top of `TangleStorage` (must-never-upgrade-live) and restored the `__gap` consumption ledger.
- **Snapshot test blind past slot 7** → pinned the tail vars (`_slashCommitmentSnapshots`, `_managerHookGasLimit`) so any future mid-layout insertion trips CI.

**Fixed (real):**
- **MANAGER_HOOK_GAS_LIMIT 500k→2M widened all-chain DoS** → reverted the flat 2M; default is back to **500k** (tight all-chain bound), with a per-deployment admin-settable `_managerHookGasLimit` (`setManagerHookGasLimit`, capped 8M). `FullDeploy` sets 2M on Tempo via `core.managerHookGasLimit`; cheap chains keep 500k. Test pins both the 500k-default bound and the raised clear.
- **bindings missing crate-root re-export** → added `IMasterBlueprintServiceManager` to `lib.rs`.
- **interface omits sibling events** → declared `BinaryVersionRecorded`/`OperatorBinaryAckRecorded` on the interface (removed concrete dups).
- **getBlueprintDefinition NatSpec overpromised** → now states display strings are dropped on-chain and the reconstruction ≠ `blueprintDefinitionHash` (which anchors the event payload).

Full suite green. The stale `TNT_CORE_VERSION` points at the gen-time HEAD (records HEAD before the commit that includes the bindings — a one-commit lag inherent to `xtask gen-bindings`).

tangletools:
## ⚠️ Review Interrupted — `3f2b9f05`

The review runner stopped before publishing a final verdict: `max_run_seconds`.

| State | Detail |
|---|---|
| Interrupted | max run seconds |

> No review verdict was produced for this run. Trigger a fresh review on the current PR head if the PR is still open.

<sub>tangletools · tangle-network/tnt-core#193 · model: `kimi-for-coding` · updated 2026-07-02T12:52:41Z</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `3f2b9f05`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T13:07:16Z</sub>

tangletools (commented):

| | |
|---|---|
| **Concerns** | 1 (1 weak-concern) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 223.9s (2 bridge agents) |
| **Total** | 224.2s |

### 💰 Value — sound

Replaces redundant on-chain blueprint-definition storage with a 32-byte keccak digest plus event-sourcing, and makes the manager-hook gas budget per-deployment configurable — a coherent, high-value gas optimization that follows existing hash+event patterns in the repo.

- **What it does:** Changes `createBlueprint` so it no longer SSTOREs the full `abi.encode(def)` blob, no longer stores 16 human-readable display strings (job name/description/metadataUri and most metadata prose fields), and no longer stores the full encoded definition in `MasterBlueprintServiceManager`. Instead it stores a single `bytes32 _blueprintDefinitionHash` on Tangle and a matching digest in the MBSM record, 
- **Goals it achieves:** Cuts `createBlueprint` gas by ~64% so it fits under the 30M block cap on high-metering chains like Tempo (where it previously OOG'd above 28M), eliminates storage of data that no on-chain invariant reads, and preserves verifiability of off-chain definition copies via the on-chain digest. The configurable hook budget keeps a tight DoS bound on cheap chains while letting high-SSTORE-metering chains 
- **Assessment:** Good change, built in the grain of the codebase. The selection of what stays on-chain vs. what moves to events is principled (schemas, sources, config, name, profilingData stay; display strings go). Tests pin job count/order/schema round-tripping, event-decode integrity, and the hash anchor (`BlueprintDefinitionStorage.t.sol:469-496`). The configurable gas budget is capped and tested (`ManagerHook
- **Better / existing approach:** none — this is the right approach. The codebase already uses the same hash+event-source pattern for TEE attestation commitments (`_requestTeeCommitmentRoot` / `_serviceTeeCommitmentRoot` / `TeeCommitmentsRecorded` in `ServicesApprovals.sol:103-144` and `TangleStorage.sol:405-417`), so this change extends an established primitive rather than reinventing one.
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

### 🎯 Usefulness — sound-with-nits

Replaces a redundant multi-KB definition blob and 16 display strings with a keccak digest + event-sourced payload; fully wired through createBlueprint, the master manager, the indexer, the Rust bindings, and the deploy script, with a configurable hook-gas budget that makes blueprint registration via

- **Integration:** End-to-end reachable. createBlueprint (src/core/BlueprintsCreate.sol:57-74) now keccak-hashes the ABI-encoded def into _blueprintDefinitionHash and forwards the full bytes to MasterBlueprintServiceManager.onBlueprintCreated, which emits BlueprintDefinitionRecorded carrying the canonical payload (src/MasterBlueprintServiceManager.sol:61-65). That event is already consumed by the indexer (indexer/sr
- **Fit with existing patterns:** Follows the codebase's existing event-sourcing grain rather than inventing a new pattern. The master manager was already the protocol-wide event sink and already emitted BlueprintDefinitionRecorded; this PR just stops ALSO SSTORE'ing the same bytes redundantly (blob + decomposed fields + MBSM record = 2-3x duplication). SchemaLib.validateJobParams/Result now takes JobDefinition directly (src/libra
- **Real-world viability:** Robust on the paths that matter. Job count/order/schemas round-trip exactly through getBlueprintDefinition (test_JobsRoundTripExactlyThroughDefinitionView). Sources reflect post-genesis mutations (test_GetBlueprintDefinition_ReflectsLiveSources), so setBlueprintSources stays load-bearing for binary resolution. The hook budget is bounded both ways: 500k default, 8M ceiling (TangleCoreApi.sol:55), h
- **Bridge attempts:** 1

### 🎯 Usefulness Audit

🟡 **blueprintDefinitionHash anchors the EVENT payload, not the view output — easy to misuse** `[ergonomics]` ``
> getBlueprintDefinition reconstructs from decomposed fields and intentionally DROPS display strings (src/core/BlueprintsManage.sol:96-118), so keccak256(abi.encode(getBlueprintDefinition(id))) will NOT equal blueprintDefinitionHash(id). The inline NatSpec documents this, and test_DefinitionHashAnchorsEventSourcedCopy pins the correct contract (hash the event payload, not the view). An off-chain verifier who hashes the view output instead of fetching+hashing the event payload will get a false mism

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T134412Z</sub>

tangletools:
## ⚠️ Review Interrupted — `3f2b9f05`

The review runner stopped before publishing a final verdict: `max_run_seconds`.

| State | Detail |
|---|---|
| Interrupted | max run seconds |

> No review verdict was produced for this run. Trigger a fresh review on the current PR head if the PR is still open.

<sub>tangletools · tangle-network/tnt-core#193 · model: `kimi-for-coding` · updated 2026-07-02T15:39:09Z</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `3f2b9f05`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T15:47:16Z</sub>

tangletools (commented):

| | |
|---|---|
| **Concerns** | 2 (2 weak-concern) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 311.3s (2 bridge agents) |
| **Total** | 311.6s |

### 💰 Value — sound-with-nits

Replaces redundant on-chain blueprint definition storage with a keccak256 digest and event-sourced payload, cutting createBlueprint gas ~64% and unblocking registration on high-metering chains; the core approach is sound and the follow-ups fixed the two original correctness gaps.

- **What it does:** Stops storing the full abi.encode(def) blob, the decomposed display strings (job name/description/metadataUri and 7 of 9 metadata prose fields), and the write-only MBSM full-payload record. Instead it stores a 32-byte _blueprintDefinitionHash, emits the full bytes in MasterBlueprintServiceManager.BlueprintDefinitionRecorded, and exposes blueprintDefinitionHash(id) so off-chain consumers can verify
- **Goals it achieves:** Make blueprint registration feasible on chains that meter storage far above mainnet (the PR cites Tempo, where createBlueprint OOG'd above 28M); eliminate storage of on-chain-unread display data; preserve verifiability of the definition via a hash anchor; keep schema validation and operator binary resolution working.
- **Assessment:** Good change on its merits. Hash-and-event-source is the right architecture for data that no trustless invariant reads, and the codebase already uses the same pattern for blueprint source sets (_blueprintSourcesHash in src/core/Base.sol:419). The PR is well tested (BlueprintDefinitionStorage.t.sol round-trips jobs, anchors the event hash, checks metrics role wiring) and the two issues in the origin
- **Better / existing approach:** For job storage, keep a minimal schema-only struct on-chain and reconstruct JobDefinition with empty display strings inside getBlueprintDefinition, rather than storing JobDefinition with three empty strings per job. On-chain callers only ever read paramsSchema/resultSchema (src/libraries/SchemaLib.sol:194-213, src/core/JobsSubmission.sol:355-363, src/core/JobsAggregation.sol:401-409), so the empty
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

### 🎯 Usefulness — sound

A well-executed storage-trim that replaces redundant on-chain blobs with a keccak digest + event-sourcing, fully wired into the existing indexer and deploy path, fitting an established pattern in the codebase.

- **Assessment:** Net positive and will do its job. The change removes genuine waste (a multi-KB `abi.encode` blob stored 2-3x at 20k gas/slot plus 16 write-only display strings) and, per the PR, unblocks `createBlueprint` on Tempo where it OOG'd past the block cap. The event-sourcing contract is verified by tests, the indexer already consumes the canonical copy, and the deploy config wires the gas budget for the h
- **Integration:** Fully reachable and wired. `createBlueprint` (BlueprintsCreate.sol:26) is the primary developer entry point and stays so. The new `blueprintDefinitionHash(id)` view (BlueprintsManage.sol:79) is exposed on `ITangleBlueprints` and bound into the Rust bindings (commit 0ea2efa). Critically, the off-chain consumer ALREADY EXISTS: the Envio indexer subscribes to `BlueprintDefinitionRecorded` at indexer/
- **Fit with existing patterns:** Fits the established grain precisely. The codebase already uses the 'store a digest, emit the full payload' pattern: `_requestTeeCommitmentRoot`/`_serviceTeeCommitmentRoot` (TangleStorage.sol:411-417) store keccak roots while `TeeCommitmentsRecorded` carries the array; `MasterBlueprintServiceManager` was ALREADY the event sink for `BlueprintDefinitionRecorded`/`BinaryVersionRecorded`/`OperatorBina
- **Real-world viability:** Holds up under realistic use. On-chain logic reads are preserved exactly: job COUNT + index ORDER + param/result schemas (load-bearing for `JobsSubmission`/`JobsAggregation`/`JobsRFQ`, all retargeted to `_jobDefinition`), `metadata.name` + `profilingData` (operator manager reads them per BlueprintsCreate.sol:110-123), `metadataUri`/`metadataHash`, sources, config, memberships. The `getBlueprintDef
- **Bridge attempts:** 1

### 💰 Value Audit

🟡 **Job definitions stored with empty display strings waste slots** `[better-architecture]` ``
> src/core/BlueprintsCreate.sol:81-98 pushes Types.JobDefinition with name/description/metadataUri set to "" so getBlueprintDefinition can return the full shape, but every on-chain consumer (SchemaLib at src/libraries/SchemaLib.sol:194-213, JobsSubmission/RFQ/Aggregation) only uses paramsSchema/resultSchema. Each empty string still occupies one storage word, so a 5-job blueprint pays for 15 slots it does not need. A smaller on-chain struct (just bytes paramsSchema / bytes resultSchema) with recons

🟡 **Large TangleCoreApi refactor is bundled but not described in the PR** `[proportion]` ``
> The PR title and body describe only hash+event-source and gas savings, yet the commit also extracts ~460 lines of admin/view functions from src/core/Base.sol into a new src/core/TangleCoreApi.sol mixin (git diff fb6a47b^..HEAD -- src/core/Base.sol src/core/TangleCoreApi.sol is ~1059 lines). The rationale given in TangleCoreApi.sol:11-17—reducing facet bytecode—is reasonable, but the refactor is orthogonal to the stated goal and makes the change harder to review than the title implies.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T155255Z</sub>

tangletools:
## ✅ No Blockers — `3f2b9f05`

**Review health 100/100** · **Reviewer score 33/100** · **Confidence 95/100** · 21 findings (3 medium, 18 low)

| | opencode-kimi | glm | aggregate |
|---|---|---|---|
| **Readiness** | 67 | 33 | **33** |
| **Correctness** | 67 | 33 | **33** |
| **Security** | 67 | 33 | **33** |
| **Testing** | 67 | 33 | **33** |
| **Architecture** | 67 | 33 | **33** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 8/8 planned shots over 49 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 49 changed files. Global verifier still owns final merge decision.

🟠 **MEDIUM** Storage layout is non-upgradeable greenfield — in-place UUPS upgrade of any existing proxy (e.g. base-sepolia) corrupts ALL state — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol)
> The diff removes _blueprintDefinitionBlobs, _blueprintJobSchemas, _pendingNativeRewards, _deprecatedTntStakerFeeBps from their original slot positions and inserts _blueprintHasConfig (slot ~after _blueprintConfigs), _blueprintDefinitionHash (replacing the blob slot), _blueprintJobs (replacing the job-schema slot) in the middle of the layout, plus appends _managerHookGasLimit (__gap 28->27). This shifts every storage slot from the first inserted mapping onward. The contract natspec at TangleStorage.sol:10-19 explicitly documents this as GREENFIELD/PRE-MAINNET and forbids using this impl as an in-place UUPS upgrade of any existing proxy. Impact: if the deploy process (script/FullDeploy.s.sol) is mistakenly pointed at an existing proxy's admin proxy-admin upgradeTo, every mapping/variable aft

🟠 **MEDIUM** Storage layout snapshot test tail assertions are non-functional — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol#L123-L134)
> The PR adds _managerHookGasLimit at the layout tail. StorageLayoutSnapshotTest.t.sol:55-56 now asserts _pinnedTangleSlot('_slashCommitmentSnapshots') == 89 and _pinnedTangleSlot('_managerHookGasLimit') == 90, but _pinnedTangleSlot() ([lines 123-134](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol#L123-L134)) only has branches for slots 0-7 and returns type(uint256).max for every other field. Those two assertions therefore always fail, so the storage-guard CI job breaks and future mid-layout insertions are no longer caught.

🟠 **MEDIUM** Public views blueprintMetadata() and getBlueprintDefinition() now return lossy display data — downstream SDK/indexers must migrate to event-sourced reads — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/core/BlueprintsCreate.sol)
> _storeBlueprintMetadata now zeroes 7 of 9 BlueprintMetadata fields (description/author/category/codeRepository/logo/website/license) keeping only name+profilingData, and _storeBlueprintSchemas pushes JobDefinition with empty name/description/metadataUri. Consequently the public views BlueprintsManage.sol:38 blueprintMetadata() and BlueprintsManage.sol:94 getBlueprintDefinition() return mostly-empty display data. I confirmed via grep that NO on-chain logic reads these dropped fields (only metadata.name + profilingData are consumed), and the full data round-trips through the BlueprintDefinitionRecorded event emitted by MasterBlueprintServiceManager.sol:64, anchored by _blueprintDefinitionHash (test_DefinitionHashAnchorsEventSourcedCopy, test_MetadataSourcesAndMembershipsPersisted, test_JobsR

🟡 **LOW** Inconsistent raw-identifier style between mod.rs and lib.rs — [`bindings/src/lib.rs`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/bindings/src/lib.rs)
> mod.rs (bindings/src/bindings/mod.rs:6-24) declares modules without r# prefixes (e.g., `pub mod i_master_blueprint_service_manager;`), while lib.rs continues to use raw-identifier paths for all re-exports (e.g., `pub use bindings::r#i_master_blueprint_service_manager::IMasterBlueprintServiceManager;`). This compiles because `r#foo` and `foo` name the same identifier, but it leaves the hand-maintained export file out of sync with the generated module declarations. Align the two files by removing the redundant `r#` prefixes in lib.rs.

🟡 **LOW** __gap accounting comment has an off-by-one (pre-existing, carried forward) — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol)
> The comment states 'Consumed from the standard 50: initial(10) + binary-versions(5) + supply-chain-hardening(5) + slash-commitment-snapshot(1) + manager-hook-gas-limit(1)' = 22 consumed, implying 50−22 = 28 remaining slots. But __gap is declared as [27]. The real starting gap was 49 (not 50): pre-PR the gap was 28 with 21 consumed (49−21=28). This is a pre-existing off-by-one in the 'standard 50' claim, carried forward into the new comment. NOT a correctness bug — verified via `forge inspect` that total storage is preserved at 118 slots (slots 0–117), so the upgrade-safety invariant holds. The comment is simply misleading about the starting count. Fix: either change 'standard 50' to 'standard 49', or add one more consumed entry (the initial block consumed 11, not 10). Recommend correcting 

🟡 **LOW** __gap comment length does not match declared array — [`src/TangleStorage.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol#L549)
> The reserved-gap comment at [line 549](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol#L549) says '__gap size 28' and the consumed-slot arithmetic sums to 22 (50-22=28), but the actual declaration at [line 553](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/TangleStorage.sol#L553) is 'uint256[27] private __gap;'. This mismatch can mislead future maintainers when decrementing the gap for appended variables and risks an off-by-one storage layout mistake on the next append.

🟡 **LOW** _hookGasLimit() is a single global knob — raising it for one high-SSTORE chain widens the DoS bound for ALL blueprint managers — [`src/core/Base.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/core/Base.sol)
> setManagerHookGasLimit (TangleCoreApi.sol) lets ADMIN_ROLE raise the per-hook budget from the 500k default up to 8M. The value is global (one _managerHookGasLimit for every blueprint manager), so raising it to clear a real BSM setup hook on a high-SSTORE-metering chain (e.g. Tempo ~11x) also lets every other blueprint's manager burn up to 8M gas per hook in _callManager/_tryCallManager/_callManagerView and in PaymentsBilling/PaymentsDistribution staticcalls. A blueprint creator who deploys a gas-guzzling manager amplifies that to 8M per invocation. The code documents this tradeoff and it is ADMIN_ROLE-gated (not attacker-controlled), but it is a meaningfully wider worst-case bill/terminate gas surface than the prior fixed 500k. Fix (optional): consider a per-blueprint or per-manager gas bu

🟡 **LOW** getBlueprintDefinition() now returns normalized config + resolved revision instead of raw creation-time inputs — [`src/core/BlueprintsManage.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/core/BlueprintsManage.sol)
> Reconstruction reads definition.config = _blueprintConfigs[blueprintId] (the NORMALIZED config — Fixed/PayOnce when hasConfig is false) and definition.masterManagerRevision = _blueprintMasterRevisions[blueprintId] (the RESOLVED revision, never 0). The old blob decode returned def.config (raw) and def.masterManagerRevision (the requested value, possibly 0 = 'latest'). So for hasConfig=false blueprints or revision-0 requests the view now returns different values than the genesis blob did. This is arguably more correct (the view reflects what the protocol actually uses), and keccak256(abi.encode(view)) is explicitly documented as NOT equal to blueprintDefinitionHash anyway (the hash anchors the full event payload), so no integrity check breaks. Flagged only as a consumer-visible semantic chan

🟡 **LOW** Manager hook gas cap ceiling exceeds documented tight bound — [`src/core/TangleCoreApi.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/core/TangleCoreApi.sol#L488-L490)
> TangleCoreApi.setManagerHookGasLimit enforces `if (limit > 8_000_000) revert Errors.InvalidState();`, but Base.sol [lines 488-490](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/core/TangleCoreApi.sol#L488-L490) state that '2M keeps the reentrancy/DoS bound tight while clearing that headroom', and ManagerHookGasBudget.t.sol:57 exercises a 2M budget. The 8M ceiling is four times looser than the documented tight DoS bound, creating a mismatch between the code's actual admin-settable maximum and the stated security model. Align the cap to 2M or update the rationale to justify 8M.

🟡 **LOW** BinaryVersionRecorded.sha256Hash unindexed limits indexer filtering — [`src/interfaces/IMasterBlueprintServiceManager.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/interfaces/IMasterBlueprintServiceManager.sol)
> The event is documented as the authoritative cross-blueprint indexer sink for binary publishes, yet `sha256Hash` (bytes32) is unindexed. Only blueprintId+versionId are indexed. Indexers/consumers that want to dedup or filter by content hash must scan all events. Making `sha256Hash indexed` (dropping the indexed slot is fine since versionId+sha256Hash would still be 2 indexed, or keep all three: blueprintId indexed + versionId indexed + sha256Hash indexed = 3, the legal max, dropping binaryUri to unindexed which it already is) would let consumers filter by hash in O(1) at the topic level. Minor design preference, not a correctness defect — the event still carries all data.

🟡 **LOW** Incidental enum whitespace edit bundled with struct removal — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/libraries/Types.sol)
> The diff deletes a trailing blank line inside the QuoteOperation enum body. Unrelated to the StoredJobSchema removal and purely cosmetic (forge fmt neutral), but mixes an orthogonal change into a focused refactor. No action required; flagging only for diff hygiene.

🟡 **LOW** Custom payment errors lack explicit test coverage — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol)
> New errors BadMsgValue, UnexpectedMsgValue, and EthTransferFailed replace string reverts, but no tests in test/rewards/ServiceFeeDistributor.t.sol or ServiceFeeDistributorStreaming.t.sol assert these selectors. Add targeted tests (e.g., distributeServiceFee with wrong msg.value, claim to a reverting recipient) to prevent silent regression if the helper is changed.

🟡 **LOW** No regression test pinning the All-vs-Fixed equivalence through the unified _distributeForRequirements — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol#L668-L687)
> The merge collapses four functions into two boolean-dispatched ones; the correctness proof rests on the fixedMode ternary selecting the right score/slash/accumulator triple ([lines 668-687](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol#L668-L687), 733-752). The existing rewards tests (test/rewards/ServiceFeeDistributor.t.sol etc.) were not modified in this PR and I could not execute them in-sandbox, so the unified path's symmetry is not pinned by a new test. Recommend adding a focused test that distributes the same amount through All and Fixed pools for an identical score setup and asserts the accumulator deltas match the pre-refactor expected values, so a future edit to the ternary cannot silently break one m

🟡 **LOW** _pendingPosition relies on callers to filter mode==0 — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol)
> _pendingPosition only branches on `mode == 1` and otherwise falls into the fixed-blueprint loop. If ever called with mode==0 it would iterate _fixedBlueprints (empty for a non-position) and return 0 - safe today because every caller (claimFor L772, _claimAllForToken L836, pendingRewards L1003, _harvestToken via _harvestAllTokens from _settleExpiredLock L1032) guards mode!=0. This matches the pre-refactor _harvestToken/pendingRewards structure, so it is not a regression; flagging only because the new helper's view signature makes it look more general than its precondition. A `if (mode == 0) return 0;` guard at the top would make the invariant explicit.

🟡 **LOW** _syncDebtsToCurrentAcc no longer zeroes _debtAll when score has collapsed to 0 — [`src/rewards/ServiceFeeDistributor.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol#L893)
> Old mode==1 path computed `_debtAll = (userScore * acc) / PRECISION` unconditionally, so a full withdrawal (userScore->0) zeroed every per-token debt entry for that (delegator,operator,assetHash). New path delegates to _settlePositionDebt which `return`s early when userScore==0 ([line 893](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol#L893)), leaving _debtAll at its last-harvested (non-zero) value, and _pruneDelegatorPosition ([line 1091](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/src/rewards/ServiceFeeDistributor.sol#L1091)) does not touch _debtAll/_debtFixed. I traced the re-delegation path: on the next onDelegationChanged, _harvestAllTokens runs with the post-pr

🟡 **LOW** _decodeRecordedDefinition assumes a single BlueprintDefinitionRecorded emission — [`test/blueprints/BlueprintDefinitionStorage.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/BlueprintDefinitionStorage.t.sol)
> The helper returns on the first topic match. All current tests pass address(0) as manager so only the master manager emits the event — safe today. But if a custom BSM ever emitted the same signature, the helper would silently pick the wrong log. Could assert exactly-one-match for robustness. Test-helper nit, not a production issue.

🟡 **LOW** Gas-budget OOG test bound (6M) is generous; depends on current SSTORE pricing — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/ManagerHookGasBudget.t.sol#L40)
> assertLt(gasBefore - gasleft(), 6_000_000) correctly pins that hostile-hook damage is bounded, but the 6M constant implicitly assumes ~20k/SSTORE for the pre-hook storage writes. If a future EVM schedule slashed SSTORE cost below ~5k, test_DefaultBudgetBoundsHeavySetupHook ([line 40](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/ManagerHookGasBudget.t.sol#L40)) could fail because 54 writes might clear 500k. Inherent to gas-metering tests; documenting so a future schedule change isn't a surprise.

🟡 **LOW** Loose gas bound in hostile-hook test — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/ManagerHookGasBudget.t.sol)
> The assertion `assertLt(gasBefore - gasleft(), 6_000_000, ...)` is ~12x the 500k default hook budget. It catches a removed budget (transaction would burn far more), but it does not tightly verify that the hook is cut off near the configured budget. A tighter bound (e.g., < 1_000_000 or < 1_500_000) would make the DoS-bound test precise enough to catch a regressed/raised effective budget.

🟡 **LOW** New ManagerHookGasBudget test file does not cover setManagerHookGasLimit access control or the 8M ceiling revert — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/ManagerHookGasBudget.t.sol)
> The file is dedicated to the configurable hook gas budget but tests only the happy/DoS paths. TangleCoreApi.sol:54-58 enforces onlyRole(ADMIN_ROLE) and reverts on limit > 8_000_000 — neither path is tested. A non-admin attempt and an over-ceiling attempt would round out the pin. Coverage gap, not a defect.

🟡 **LOW** No test that setManagerHookGasLimit(0) resets to the 500k default after an override — [`test/blueprints/ManagerHookGasBudget.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/ManagerHookGasBudget.t.sol)
> TangleCoreApi.sol:50-52 and Base.sol:499-501 document that zero 'resets to the 500k default.' The tests check default==500k and raised==2M but never exercise set→non-zero→set(0)→verify-500k. Minor coverage gap for the documented reset semantics.

🟡 **LOW** Compiler warning on view test helper — [`test/blueprints/MasterBlueprintServiceManager.t.sol`](https://github.com/tangle-network/tnt-core/blob/3f2b9f05f2e886af4f0b3d080b6c29cf895b62e7/test/blueprints/MasterBlueprintServiceManager.t.sol)
> Foundry emits Warning 2018: function state mutability can be restricted to `view`. Trivial cleanup; does not affect correctness.

---
<sub>tangletools · 2026-07-02T17:06:49Z · [trace](https://gist.github.com/tangletools/57245bf9e6338622895c1b731e4a5386)</sub>

tangletools (approved):
## ✅ Approved — 21 non-blocking findings — `3f2b9f05`

> Full multi-shot audit completed 8/8 planned shots over 49 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 49 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/57245bf9e6338622895c1b731e4a5386)

Summary comment for this run: [full summary](https://github.com/tangle-network/tnt-core/pull/193#issuecomment-4868432804)

---
<sub>tangletools · 2026-07-02T17:06:49Z · immutable [trace](https://gist.github.com/tangletools/57245bf9e6338622895c1b731e4a5386)</sub>

PR #194, opened 2026-07-02
feat(operators): event-source rpcAddress + derive operatorCount (lifecycle gas, stacked on #193)

drewstone:
Stacked on **#193**. Applies the same store-hash/emit-event + de-duplication discipline to the next-heaviest lifecycle writers, after an adversarial audit that **kept the economic core untouched**.

## Cuts (safe — no invariant reads the removed data)

**1. Operator `rpcAddress` → event-only.** `registerOperator` / `updateOperatorPreferences` no longer SSTORE the RPC discovery string; it's carried in full by `OperatorRegistered` / `OperatorPreferencesUpdated` and the BSM hook payload (the indexer already reads it from the event). `ecdsaPublicKey` stays on-chain (read for gossip-key dedup + the proof-of-possession front-run defense). Removes one unbounded-string SSTORE per register/update — **~120–220k gas/register on Tempo** (6–11× SSTORE). `getOperatorPreferences().rpcAddress` now returns `""` (documented).

**2. `Blueprint.operatorCount` counter → derived.** Dropped the cached counter (and its `++`/`--` SSTOREs on register/unregister) that duplicated `_blueprintOperators` (EnumerableSet); `blueprintOperatorCount(id)` already returns `.length()`. Removes a drift-bug class + a warm SSTORE per join/leave. **ABI:** `getBlueprint()` drops `operatorCount`; bindings regenerated. No in-repo reader used the struct field (only `blueprintOperatorCount()`).

## Deliberately NOT cut (audited, kept)
- **Security-commitment representations** (`_requestSecurityCommitments` / `_serviceSecurityCommitments` / `_serviceSecurityCommitmentBps`): the arrays are **not** view-only — they're iterated by slashing, billing, and distribution. Cutting any would touch economic accounting. Kept.
- **Denormalized `membership`/`pricing`** on the `Blueprint` struct: frees no storage slot (packing), so ~0 gas — pure ABI churn. Deferred.

## Verify
- `ci_size` build green; bindings regenerated (Blueprint tuple + `blueprintOperatorCount` derive).
- Full suite green (operator lifecycle, front-run / proof-of-possession, BSM-hook payload, payments, slashing).
- SDK impact: bindings-only (no logic change); the operator manager reads `rpcAddress` from the event already.


tangletools (approved):
## ✅ Auto-approved drewstone PR — `7848aa07`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T07:00:55Z</sub>

tangletools (commented):
## 🟢 Value Audit — sound

| | |
|---|---|
| **Verdict** | sound |
| **Concerns** | 2 (2 weak-concern) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 328.3s (2 bridge agents) |
| **Total** | 328.6s |

### 💰 Value — sound

Removes two redundant storage writes from operator lifecycle by event-sourcing rpcAddress and deriving blueprint operator count from the existing EnumerableSet, saving lifecycle gas without breaking on-chain invariants.

- **What it does:** Stops persisting operator rpcAddress on-chain (it is now carried only by OperatorRegistered/OperatorPreferencesUpdated events and the BSM hook payload) and removes the cached Blueprint.operatorCount field, deriving the count from _blueprintOperators[blueprintId].length() via the existing blueprintOperatorCount() view.
- **Goals it achieves:** Cuts warm SSTOREs on registerOperator, unregisterOperator, and updateOperatorPreferences; eliminates a denormalized counter that duplicated the canonical membership set; extends the same hash+event-source discipline introduced in #193 to the next heaviest lifecycle writers while keeping the economic core untouched.
- **Assessment:** Coherent and safe. No on-chain invariant reads rpcAddress from storage (gossip dedup uses ecdsaPublicKey), and blueprintOperatorCount() already exposed the canonical set length. The diff updates the one in-repo test that read bp.operatorCount and regenerates bindings; comments explicitly document the new event-source contract.
- **Better / existing approach:** none — this is the right approach. The codebase already uses blueprintOperatorCount() as the canonical count source and just applied the same store-hash/emit-event pattern from #193 (Blueprint definition) to operator metadata. Removing rpcAddress from the struct entirely would break the ABI more than this gas-saving, compatibility-preserving choice.
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

### 🎯 Usefulness — sound

Event-sources the operator RPC string and derives the blueprint operator count from the membership set — both cuts remove redundant SSTOREs with every consumer verified reachable, and follow the event-sourcing discipline established in #193.

- **Integration:** Fully reachable. rpcAddress is still accepted as a param, emitted in OperatorRegistered/OperatorPreferencesUpdated (src/core/Operators.sol:341,456), and forwarded to the BSM hook with the freshly-supplied value (Operators.sol:346,446) — never read from storage by any on-chain path. The indexer already consumes it from the events (indexer/src/handlers/tangle.ts:488 for register, :526 for update), s
- **Fit with existing patterns:** Exact fit. This is the second application of the store-hash/emit-event + de-dup discipline introduced in #193 (blueprint-definition hashing). It keeps on-chain-invariant-relevant data (ecdsaPublicKey, read for gossip-key dedup + proof-of-possession defense at Operators.sol:_blueprintOperatorKeys) and moves only off-chain-discovery data (rpcAddress) to event-only. The Blueprint.operatorCount counte
- **Real-world viability:** Holds up. Deriving count from the set's .length() removes an entire drift-bug class (counter could never desync). The metadata-lock gate switched from bp.operatorCount==1 to _blueprintOperators[id].length()==1 after .add() (Operators.sol:328-331) — functionally equivalent, correct ordering. One documented behavior delta: updateOperatorPreferences no longer treats an empty rpcAddress as 'keep uncha
- **Bridge attempts:** 1

### 💰 Value Audit

🟡 **Stale interface doc for OperatorPreferencesUpdated rpcAddress** `[maintenance]` ``
> src/interfaces/ITangleOperators.sol:31 says rpcAddress 'may be empty if unchanged', but the value is never persisted; empty now means the caller supplied an empty string, not 'unchanged'. The implementation comment in Operators.sol:438-441 is correct, so this interface comment should be aligned in a follow-up.

### 🎯 Usefulness Audit

🟡 **updateOperatorPreferences silently advertises empty rpcAddress if caller passes empty string** `[ergonomics]` ``
> Previously `if (bytes(rpcAddress).length > 0) prefs.rpcAddress = rpcAddress` let a caller change only the ECDSA key while preserving the stored endpoint; now storage is always empty and the event/BSM-hook emit exactly the passed value (Operators.sol:441-446). A client updating only its key and passing "" will wipe its advertised endpoint from the indexer and any BSM manager until the next non-empty call. Documented in NatSpec ('pass the current endpoint on every update'), so not a blocker — but 

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T080714Z</sub>

tangletools:
## ❌ Needs Work — `7848aa07`

**Review health 100/100** · **Reviewer score 12/100** · **Confidence 90/100** · 22 findings (2 critical, 1 high, 6 medium, 13 low)

| | opencode-kimi | glm | deepseek | aggregate |
|---|---|---|---|---|
| **Readiness** | 12 | 50 | 29 | **12** |
| **Confidence** | 90 | 90 | 90 | **90** |
| **Correctness** | 12 | 50 | 29 | **12** |
| **Security** | 12 | 50 | 29 | **12** |
| **Testing** | 12 | 50 | 29 | **12** |
| **Architecture** | 12 | 50 | 29 | **12** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision.

### Blocking

🟣 **CRITICAL** Blueprint struct layout change corrupts existing storage on UUPS upgrade — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol#L37-L44)
> Removing `uint32 operatorCount` from `Blueprint` ([line 37-44](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol#L37-L44)) changes slot-2 packing from [createdAt:8][operatorCount:4][membership:1][pricing:1][active:1] to [createdAt:8][membership:1][pricing:1][active:1]. Existing stored blueprints will read membership/pricing/active from the old operatorCount bytes, not from the bytes where they were actually written. Impact: bp.active read at Operators.sol:204,257, ServicesRequests.sol:442, QuotesCreate.sol:164, QuotesExtend.sol:213; bp.membership read at ServicesRequests.sol:444, QuotesCreate.sol:234; bp.pricing read at ServicesRequests.sol:445, QuotesCreate.sol:67,84,235, TangleServicesFacet.sol:60,111. A corrupted active=fals

🟣 **CRITICAL** Blueprint struct storage layout break on upgrade — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol)
> Removing `uint32 operatorCount` from between `uint64 createdAt` and `MembershipModel membership` changes the packing of slot 2. Old layout: createdAt (bytes 0-7), operatorCount (bytes 8-11), membership (byte 12), pricing (byte 13), active (byte 14). New layout: createdAt (bytes 0-7), membership (byte 8), pricing (byte 9), active (byte 10). Because `_blueprints[blueprintId]` is stored in `TangleStorage.sol:91` and the base commit actively wrote `bp.operatorCount` in `Operators.sol`, existing blueprints will have stale operatorCount bytes interpreted as membership/pricing/active. This corrupts blueprint state after a UUPS upgrade. Either retain `operatorCount` as a deprecated unused field to preserve layout, or add an upgrade migration that rewrites every existing Blueprint before the new im

🔴 **HIGH** Blueprint struct storage layout break on upgrade — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/BlueprintsCreate.sol)
> The PR removes `operatorCount` from `Types.Blueprint` (base layout: `createdAt(8) + operatorCount(4) + membership(1) + pricing(1) + active(1)`; head layout removes the 4-byte field). `_blueprints` is a mapping in an upgradeable UUPS proxy (`TangleStorage.sol`), so every existing blueprint record written with the old layout will be mis-decoded after upgrade: the old `operatorCount` bits become `membership`/`pricing`/`active`, corrupting blueprint configuration and activation state. There is no migration or storage-gap reserve. Fix: keep the unused `operatorCount` field (or an explicit 4-byte padding slot) in `Types.Blueprint`, or provide an upgrade function that rewrites every existing `_blueprints[id]` to the new layout.

### Other

🟠 **MEDIUM** TNT_CORE_VERSION records PR base instead of HEAD despite source changes — [`bindings/TNT_CORE_VERSION`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/TNT_CORE_VERSION)
> File content is `0ea2efa13471014ceeab9dce545e10b4d0ef6e6f` (the PR base), but `git rev-parse HEAD` on this branch is `7848aa0739ced1e9f70e8ab758eb100d64886935`. The PR also changes `src/core/Operators.sol`, `src/libraries/Types.sol`, and related generated bindings (e.g., `bindings/src/bindings/i_tangle.rs` removes `Blueprint.operatorCount`), so the bindings were generated from PR-head sources. xtask/README.md states that `cargo xtask gen-bindings` 'Records the git commit hash in bindings/TNT_CORE_VERSION' and bindings/README.md exposes it as the source commit. A stale base commit misleads consumers about which Solidity source the crate was built from and breaks the version-tracking contract. Fix by running `cargo xtask gen-bindings` on the PR branch and committing the resulting `bindings/T

🟠 **MEDIUM** Breaking return-shape change: getBlueprint drops operatorCount from returned tuple — [`bindings/abi/ITangle.json`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/abi/ITangle.json)
> getBlueprint's output[0].components went from 7 to 6: the `operatorCount` (uint32, internalType uint32) field was removed from Types.Blueprint. The identical change appears in bindings/abi/ITangleFull.json (abi[38]). Any off-chain consumer decoding the return tuple positionally (indexer, Rust bindings, frontends, subgraph handlers) will misparse the remaining fields (membership/pricing/active shift down one slot) or revert on ABI-decode length mismatch. The PR does regenerate the Rust bindings (bindings/src/bindings/i_tangle.rs, i_tangle_full.rs) so internal consumers are consistent, but external consumers pinned to the prior ABI break silently. Impact: integration breakage for downstream indexers/SDKs. Fix: bump the bindings crate major version and surface the breaking change in the chang

🟠 **MEDIUM** Existing operator RPC addresses not cleared or migrated — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol)
> New registrations now store `rpcAddress: ""`, and `updateOperatorPreferences` no longer writes `rpcAddress` to storage. However, pre-existing `_operatorPreferences[blueprintId][op].rpcAddress` values remain in storage after an upgrade. `TangleCoreApi.getOperatorPreferences` (src/core/TangleCoreApi.sol:431) and its natspec claim the returned `rpcAddress` is always empty, but for existing operators the view will return the stale stored value until they unregister. This creates an inconsistent API across migrated and new operators. Fix: clear `rpcAddress` on the first post-upgrade update/unregister, or have `getOperatorPreferences` return a struct with `rpcAddress: ""` without reading the stored string.

🟠 **MEDIUM** updateOperatorPreferences silently broadcasts empty rpcAddress when caller passes "" — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol#L399-L401)
> Old behavior: `if (bytes(rpcAddress).length > 0) prefs.rpcAddress = rpcAddress;` — empty meant 'keep stored value'. New behavior: `abi.encode(... rpcAddress: rpcAddress)` and `emit OperatorPreferencesUpdated(..., rpcAddress)` — empty is forwarded as-is. An operator who wants to change ONLY their ecdsaPublicKey (a common case — key rotation) and passes `""` for rpcAddress per the previously-documented contract will now broadcast an empty endpoint in the event and to the BSM hook, silently breaking their gossip reachability until they re-call with the real endpoint. The contract-internal doc ([lines 399-401](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol#L399-L401)) says 'pass the current endpoint on every update', but the publ

🟠 **MEDIUM** Removing uint32 operatorCount from Blueprint struct shifts storage layout of remaining fields — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol)
> The `Blueprint` struct is stored in the `_blueprints[blueprintId]` mapping (TangleStorage.sol). Previously slot 2 packed: createdAt(8) + operatorCount(4) + membership(1) + pricing(1) + active(1) = 15 bytes. After removal: createdAt(8) + membership(1) + pricing(1) + active(1) = 11 bytes. The fields `membership`, `pricing`, `active` move from byte offsets 12/13/14 to 8/9/10 within that slot. For UUPS-upgradeable contracts (per CLAUDE.md all core contracts are upgradeable), upgrading a live deployment to this code would cause `membership` to read stale bytes from the old `operatorCount` high byte, `pricing` to read old `membership`, and `active` to read old `pricing` — silently corrupting every existing blueprint. Impact: broken service-request validation, wrong pricing model, wrong active fl

🟠 **MEDIUM** rpcAddress on-chain view now returns empty for all operators — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol)
> registerOperator now persists rpcAddress as '' (Operators.sol:319) and updateOperatorPreferences no longer writes it to storage (Operators.sol:438-441). getOperatorPreferences().rpcAddress (TangleCoreApi.sol:418-422) returns whatever is in storage — which for post-upgrade registrations/updates will be empty, but for pre-existing operators it still returns the old stored value until overwritten. Off-chain consumers reading the view instead of OperatorRegistered/OperatorPreferencesUpdated events will see stale or empty values. Docstrings updated to declare rpcAddress as event-only — contractually consistent, but an upgrade-time behavioral change. No security impact (ecdsaPublicKey, the on-chain invariant, is preserved).

🟡 **LOW** TNT_CORE_VERSION under-reports actual binding state — [`bindings/TNT_CORE_VERSION`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/TNT_CORE_VERSION)
> The version file was bumped from fb6a47b to 0ea2efa, but the binding source files (i_tangle.rs, i_tangle_blueprints.rs, i_tangle_full.rs) were further manually edited at commit 7848aa0 to remove `operatorCount` from the Blueprint struct, matching Solidity changes at 7848aa0. The version hash should reflect the commit whose contracts the bindings are compatible with. Pointing to 0ea2efa (which still has `operatorCount` in Blueprint) will mislead consumers who use the version to select which contracts to deploy against. Fix: either regenerate bindings via `cargo xtask gen-bindings` at HEAD so the version becomes 7848aa0, or accept the manual-edits pattern and bump to the latest commit that modified contracts.

🟡 **LOW** Breaking ABI change: getBlueprint output struct shape changed — [`bindings/abi/ITangle.json`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/abi/ITangle.json)
> The `operatorCount` field was removed from `Types.Blueprint` returned by `getBlueprint`. Any off-chain client decoding `getBlueprint` outputs with the old ABI will get malformed data. The replacement is `blueprintOperatorCount(blueprintId)`. This is pre-mainnet, so impact is limited, but downstream consumers (indexer, SDKs, frontend) must update their ABI decoders. Test at test/Tangle.t.sol:48 was updated to use the new function.

🟡 **LOW** Same getBlueprint output change in ITangleFull.json — [`bindings/abi/ITangleFull.json`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/abi/ITangleFull.json)
> Mirrors the same `operatorCount` removal from `Types.Blueprint` in the full contract ABI. Consistent with ITangle.json.

🟡 **LOW** Regenerated ABI artifacts carry no authoring defects, but storage-layout safety must be confirmed at source — [`bindings/abi/MultiAssetDelegation.json`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/abi/MultiAssetDelegation.json)
> Confirmed via deep diff: IBlueprintServiceManager, IMultiAssetDelegation, and MultiAssetDelegation have byte-identical `abi` arrays to base — their only delta is in the embedded `metadata` (source keccak hashes / NatSpec), which is cosmetic and carries no interface or selector change. methodIdentifiers are unchanged across all five files. The ABI change to Blueprint is therefore fully attributable to the Types.sol struct edit and is accurately propagated. The one cross-cutting risk the ABI cannot self-validate: removing a packed field from an in-storage struct in a UUPS-upgradeable contract re-packs slot bytes (membership/pricing/active move from bytes 12-14 to bytes 8-10), so a live upgrade would read garbage for those fields absent a storage migration. That belongs to the Types.sol shot;

🟡 **LOW** Generated bindings not compile-verified in this shot — [`bindings/src/bindings/i_tangle.rs`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/bindings/src/bindings/i_tangle.rs)
> cargo build -p tnt-core-bindings could not run (rustc 1.88.0 ICE on transitive deps once_cell/heck/serde_json/paste in sandbox). Not a code defect — the diff is mechanical and self-consistent by inspection — but the global verifier should confirm a green bindings build in a working toolchain before merge, since downstream Rust consumers indexing Blueprint.{membership,pricing,active} by position would break at compile time (expected and intended for a struct-field removal).

🟡 **LOW** Storage-layout note for the global verifier: Blueprint struct field removed mid-struct — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/BlueprintsCreate.sol)
> The BlueprintsCreate.sol change (dropping `operatorCount: 0` from the struct literal) compiles only because Types.sol (out of shot) removed `uint32 operatorCount` from the middle of `struct Blueprint`. That shifts byte offsets within slot 2 of every `_blueprints[id]` value: `membership` moves from byte 12 to byte 8, `pricing` from byte 13 to byte 9, `active` from byte 14 to byte 10. For a UUPS proxy upgraded in-place, this would silently misread `membership`/`pricing`/`active` on every existing blueprint. deploy/LAUNCH-READINESS.md states the protocol is NOT yet mainnet-deployed and testnet rehearsal is still pending, so this is acceptable NOW — but this is a one-way door: the PR must not land after the first persistent (non-throwaway) deployment without an explicit storage migration. Surf

🟡 **LOW** Event emitted after external call in updateOperatorPreferences (CEI) — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol)
> The OperatorPreferencesUpdated event (L456) is emitted AFTER the external _tryCallManager call (L449-454). Compare with _registerOperator where the OperatorRegistered event (L341) is emitted BEFORE the external _callManager hook (L344-349) — the established CEI pattern. _tryCallManager does not revert (L504-509), so the event always fires. The practical risk is minimal: any reentrant call from the hook would modify the same caller's own preferences, and the event at L456 would read the reentrant-modified storage, reflecting final state. Still, the ordering is inconsistent with the codebase's own convention. Move the emit above the manager hook block for consistency.

🟡 **LOW** First-join lock check now depends on EnumerableSet.add always increasing length by exactly 1 — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol#L273)
> Replaced `bp.operatorCount++ ; if (bp.operatorCount == 1 && !locked)` with `_blueprintOperators[blueprintId].add(msg.sender); if (_blueprintOperators[blueprintId].length() == 1 && !locked)`. Behavior is equivalent for the standard path because `EnumerableSet.add` returns true and increments length by exactly 1 only when the element was absent, and the function already guards against re-registration at [line 273](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol#L273) (`OperatorAlreadyRegistered`). I verified the lock is not reset on unregister, so re-registration after a full drain does not re-emit `BlueprintMetadataLocked`. Equivalent to the old semantics. Flagging only as a dependency on an invariant (add-then-length) that is 

🟡 **LOW** Missing nonReentrant on updateOperatorPreferences — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol)
> registerOperator (L228,L242) and unregisterOperator (L354) both carry the nonReentrant modifier. updateOperatorPreferences makes an external call (L449-454, _tryCallManager to bp.manager) but has no reentrancy guard. A malicious BSM manager that re-enters updateOperatorPreferences could double-update the timestamp and key mappings within the same transaction — same-caller, same-blueprint, so no cross-operator corruption. Pre-existing gap, not introduced by this PR, but visible in the file under review. Add nonReentrant to match the pattern set by the other two operator-state-modifying functions.

🟡 **LOW** No test covers the new rpcAddress event-sourcing invariant — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol)
> The PR's central claim is that rpcAddress is event-sourced instead of storage-sourced. There is no test in the changed test surface (only test/Tangle.t.sol:48 was tweaked for `operatorCount`) that asserts: (1) `getOperatorPreferences(id, op).rpcAddress == ""` after a non-empty-rpcAddress registration; (2) the `OperatorRegistered` event carries the supplied rpcAddress in full; (3) the BSM `onRegister` payload contains the supplied rpcAddress. BSMRHooksLifecycle.t.sol:80 registers with `""`, so it cannot distinguish old from new behavior — the test passes identically on both code revisions, proving nothing about the new invariant. Without these tests, a future refactor could re-introduce storage of rpcAddress (or drop it from the event) and CI would not catch it. Fix: add a test that registe

🟡 **LOW** updateOperatorPreferences empty rpcAddress now advertises an empty endpoint — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/Operators.sol)
> Previously, passing `rpcAddress = ""` kept the stored endpoint and emitted the stored value. After this change, an empty `rpcAddress` emits `""` and forwards `""` to the BSM `onUpdatePreferences` hook. Off-chain indexers or BSMs that relied on the event/hook to discover the operator's endpoint will lose it if the caller omits the field. The updated natspec tells callers to resupply the endpoint, but this is a breaking behavioral change that should be called out in release notes or guarded with a non-empty check if an empty value is never meant to delete the endpoint.

🟡 **LOW** Doc overclaims rpcAddress is always empty — [`src/core/TangleCoreApi.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/core/TangleCoreApi.sol)
> The @dev comment says 'The returned rpcAddress is always empty: it is not persisted on-chain.' Operators registered before this upgrade still have a stored rpcAddress (the old registration/update wrote it; this PR does not clear existing storage, and unregister is the only code path that deletes the struct wholesale). For pre-upgrade operators that have never unregistered, getOperatorPreferences returns the old value, contradicting the doc. No on-chain reader uses rpcAddress from storage (verified: only getOperatorPreferences and getOperatorPublicKey read _operatorPreferences, and the latter only reads ecdsaPublicKey), so no security impact — but off-chain indexers/tooling trusting this doc could produce stale data. Fix: qualify the comment with 'for registrations after this upgrade' or ad

🟡 **LOW** Struct comment slot math uses bytes, not bits — cosmetic — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7848aa0739ced1e9f70e8ab758eb100d64886935/src/libraries/Types.sol)
> The @dev comment says 'Slot 2: createdAt (8) + membership (1) + pricing (1) + active (1) = 11 bytes'. Solidity storage slots are 32 bytes, and these are byte counts, not slot counts. The arithmetic is correct (8+1+1+1=11 bytes packed into one 32-byte slot), but '11 bytes' could be misread as '11 slots' by an auditor unfamiliar with the convention. The old comment was similarly ambiguous. No functional issue.

---
<sub>tangletools · 2026-07-02T08:40:56Z · [trace](https://gist.github.com/tangletools/6347600c02fba35d2119db8167f9c7f5)</sub>

tangletools (dismissed):
## ❌ 3 Blocking Findings — `7848aa07`

> Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 6/6 planned shots over 14 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/6347600c02fba35d2119db8167f9c7f5)

Summary comment for this run: [full summary](https://github.com/tangle-network/tnt-core/pull/194#issuecomment-4863839628)

---
<sub>tangletools · 2026-07-02T08:40:56Z · immutable [trace](https://gist.github.com/tangletools/6347600c02fba35d2119db8167f9c7f5)</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `c88d239a`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T11:58:12Z</sub>

drewstone:
### Audit round addressed (`c88d239`, rebased on updated #193)

**Reconciled (greenfield):** the CRITICAL/HIGH `Blueprint` struct-layout findings (removing `operatorCount` re-packs slot 2). Pre-mainnet greenfield — redeploy-only, never an in-place UUPS upgrade — so the re-pack is intended. #193 now carries the prominent `TangleStorage` greenfield warning + tail-pinned snapshot test that guard exactly this.

**Fixed (real):**
- **updateOperatorPreferences broadcast empty rpc when caller passed "" (MEDIUM)** → now reverts `OperatorRpcAddressRequired` on empty rpc. With rpc event-only there's no stored value to "keep unchanged", so the operator must supply their current endpoint every update — kills the silent-empty-endpoint footgun.
- **CEI: event after external hook (LOW)** → emit `OperatorPreferencesUpdated` before the manager hook (parity with `_registerOperator`).
- **stale rpc on existing operators (MEDIUM)** → greenfield: no pre-existing operators on a fresh deploy; getOperatorPreferences documents rpc as event-only.

Regenerated bindings reflect `getBlueprint` dropping `operatorCount` (consumers use `blueprintOperatorCount(id)`). Verified: OperatorLifecycle 34, Operators audit 10, Tangle.t 54, blueprints 97 green (the full test-graph compile deadlocks under via-IR — a toolchain flake, sharded per-file to prove it).

tangletools (commented):
## 🟢 Value Audit — sound

| | |
|---|---|
| **Verdict** | sound |
| **Concerns** | 1 (1 weak-concern) |
| **Heuristic** | 0.2s |
| **Duplication** | 0.1s |
| **Interrogation** | 1029.3s (2 bridge agents) |
| **Total** | 1029.6s |

### 💰 Value — sound

Removes a denormalized operator counter and an unread on-chain RPC string, deriving both from existing canonical sources — a clean, in-grain gas win with no economic-surface changes.

- **What it does:** Two storage cuts in the operator/blueprint lifecycle: (1) `OperatorPreferences.rpcAddress` is no longer SSTORE'd — it's set to "" in storage and carried solely by the `OperatorRegistered`/`OperatorPreferencesUpdated` events and the BSM hook payload (Operators.sol:319,341,346,451,454); `updateOperatorPreferences` now requires a non-empty rpcAddress since storage is no longer a fallback (Operators.s
- **Goals it achieves:** Cut gas on the highest-frequency lifecycle writers (operator register/update) by eliminating a redundant warm SSTORE per join/leave and an unbounded-string SSTORE per registration, and remove a drift-bug class (cached counter vs. canonical EnumerableSet) — the same store-hash/emit-event + de-dup discipline applied in #193, extended to the next lifecycle writers without touching economic accounting
- **Assessment:** Coherent and well-scoped. The audit correctly distinguishes read-on-chain data (ecdsaPublicKey stays — used for gossip-key dedup at Operators.sol; the Service/ServiceRequest operatorCounts stay — iterated by slashing/billing/lifecycle) from event-only data (rpcAddress, Blueprint.operatorCount). The change reuses the canonical membership set that already existed (`_blueprintOperators`) rather than 
- **Better / existing approach:** none — this is the right approach. Looked for: (a) whether a derive-from-set helper already existed — `blueprintOperatorCount()` at TangleCoreApi.sol:485 is exactly that and is now the single reader; (b) whether the permanently-empty `rpcAddress` storage field should instead be dropped from the `OperatorPreferences` struct entirely — it's kept deliberately to preserve the BSM hook's `abi.encode(Op
- **Bridge attempts:** 2
- **Bridge warning:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content

### 🎯 Usefulness — error

usefulness agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

### 💰 Value Audit

🟡 **Permanently-empty `rpcAddress` field kept in storage is a mild readability smell** `[maintenance]` ``
> `OperatorPreferences.rpcAddress` is now always "" in storage (Operators.sol:319) yet remains in the struct and is abi-encoded into the BSM hook payload (Operators.sol:451). This is justified for BSM ABI stability and is thoroughly commented, but a future reader must internalize that the field is vestigial in storage. Not blocking; consider dropping the field from the struct in a coordinated BSM-breaking change once the event-sourcing pattern is fully validated, so storage and truth align.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T131008Z</sub>

tangletools:
## ✅ No Blockers — `c88d239a`

**Review health 100/100** · **Reviewer score 60/100** · **Confidence 95/100** · 15 findings (3 medium, 12 low)

| | opencode-kimi | glm | aggregate |
|---|---|---|---|
| **Readiness** | 60 | 64 | **60** |
| **Correctness** | 60 | 64 | **60** |
| **Security** | 60 | 64 | **60** |
| **Testing** | 60 | 64 | **60** |
| **Architecture** | 60 | 64 | **60** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision.

🟠 **MEDIUM** Removing operatorCount from packed Blueprint struct shifts byte offsets of membership/pricing/active — unsafe for in-place proxy upgrade — [`src/core/BlueprintsCreate.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/BlueprintsCreate.sol#L40)
> The struct literal at [line 40](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/BlueprintsCreate.sol#L40) dropped `operatorCount: 0` because the field was removed from `Types.Blueprint`. The struct packs owner(20)+manager(20) into slots 0-1, then createdAt(8)+membership(1)+pricing(1)+active(1) into slot 2. BEFORE removal, slot 2 was createdAt(8)+operatorCount(4)+membership(1)+pricing(1)+active(1), so `membership` sat at byte offset 12, `pricing` at 13, `active` at 14. AFTER removal, `membership` reads from byte 8, `pricing` from byte 9, `active` from byte 10. For any existing proxy deployment, an in-place UUPS upgrade would cause `bp.membership`/`bp.pricing`/`bp.active` to read the stale `operatorCount` bytes instead of their real values (c

🟠 **MEDIUM** registerOperator allows empty rpcAddress, breaking the no-empty-endpoint invariant — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol#L341)
> _registerOperator copies rpcAddress to memory and emits it in OperatorRegistered ([line 341](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol#L341)) without length check. updateOperatorPreferences now reverts on empty rpcAddress ([line 417](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol#L417)) because the field is event-only and empty would break gossip reachability, but the same reasoning applies to the registration event. An operator can register with "" and publish an empty endpoint to indexers that source rpcAddress from OperatorRegistered. Fix: require bytes(rpcAddress).length > 0 in _registerOperator, or document why registration is exempt and ensure indexer

🟠 **MEDIUM** Blueprint struct field removal breaks in-storage layout for upgrades — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/libraries/Types.sol)
> Removing `uint32 operatorCount` from `Blueprint` changes the packing of slot 2. Old layout: createdAt(8) + operatorCount(4) + membership(1) + pricing(1) + active(1). New layout: createdAt(8) + membership(1) + pricing(1) + active(1). For any existing stored blueprint, `membership`, `pricing`, and especially `active` would be read from the bytes that previously held `operatorCount`, corrupting the blueprint state after an in-place upgrade. TangleStorage declares this is a greenfield/pre-mainnet layout, but the type definition itself does not enforce that; an accidental UUPS upgrade of an existing proxy would brick stored blueprints. Fix: preserve layout by keeping a deprecated `uint32 __deprecatedOperatorCount` field (or adding explicit reserved padding) until the layout freezes, and/or add 

🟡 **LOW** Pre-PR TNT_CORE_VERSION hash was not a valid repo object — [`bindings/TNT_CORE_VERSION`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/bindings/TNT_CORE_VERSION)
> The base value 69cffe5fd36ac001e8925b0ed634eb567771e284 is not a git object in this repository (git cat-file -t returns 'not an object' and no branch contains it). This violates the file's stated purpose of tracking the source commit used to generate bindings. The PR replaces it with cd28a455a4b76982a6acfbc68659a1e711c25569, a valid ancestor that matches the regenerated bindings, so the invariant is restored.

🟡 **LOW** Breaking ABI change: getBlueprint return struct drops operatorCount field — [`bindings/abi/ITangle.json`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/bindings/abi/ITangle.json)
> The 'Types.Blueprint' return tuple of getBlueprint(uint64) loses its 'operatorCount (uint32)' component between base and head (same change in ITangleFull.json). Function selector is unchanged (getBlueprint(uint64)), but the return-tuple layout shifts: membership 4->3, pricing 5->4, active 6->5. Any OFF-CHAIN consumer still pinned to the previous ABI will mis-decode the return value (silent field drift, not a revert). Impact is contained because the PR regenerates the Rust bindings (i_tangle.rs / i_tangle_full.rs / i_tangle_blueprints.rs) in-lockstep, so first-party consumers are aligned. Risk remains only for third-party/indexer pipelines that vendor the old ABI JSON. No action required IN these ABI files (they correctly reflect the new source); noting for release-notes awareness.

🟡 **LOW** ABI/EIP712 breaking change in Blueprint type — [`bindings/src/bindings/i_tangle.rs`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/bindings/src/bindings/i_tangle.rs)
> The generated EIP712 type string changed from "Blueprint(address owner,address manager,uint64 createdAt,uint32 operatorCount,uint8 membership,uint8 pricing,bool active)" to "Blueprint(address owner,address manager,uint64 createdAt,uint8 membership,uint8 pricing,bool active)" and the ABI-encoded tuple lost the operatorCount slot. This is an intentional, source-level breaking change and the bindings correctly mirror it, but any downstream Rust/TypeScript signer or decoder that still expects the old 7-field Blueprint layout will fail. No defect in the bindings themselves.

🟡 **LOW** Breaking API change for downstream Rust consumers of Blueprint.operatorCount — [`bindings/src/bindings/i_tangle.rs`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/bindings/src/bindings/i_tangle.rs)
> The `operatorCount: u32` field is removed from the `Blueprint` struct across i_tangle.rs, i_tangle_blueprints.rs, and i_tangle_full.rs. Any external crate reading `blueprint.operatorCount` will fail to compile. This is a faithful projection of the Solidity change (Types.sol:37 drops the field; count is now derived via `blueprintOperatorCount` from `_blueprintOperators[id].length()`), so the bindings are correct — but downstream consumers must migrate to the new getter. No action needed in this shot's files; flagging for the global verifier to confirm a migration note/changelog exists at the PR level.

🟡 **LOW** Missing test for OperatorRpcAddressRequired revert — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol#L417)
> The new require-empty-rpcAddress revert path introduced at [line 417](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol#L417) has no test coverage (grep found zero OperatorRpcAddressRequired hits under test/). Existing tests pass non-empty rpcAddress (e.g., test/Tangle.t.sol:228, test/audit/medlow/Operators.t.sol:184,198). Fix: add a test that updateOperatorPreferences(blueprintId, "", "") reverts with Errors.OperatorRpcAddressRequired().

🟡 **LOW** New OperatorRpcAddressRequired revert path has no test — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol)
> The PR adds `if (bytes(rpcAddress).length == 0) revert Errors.OperatorRpcAddressRequired();` — a deliberate breaking API change (rpcAddress is no longer persisted, so omitting it would broadcast an empty endpoint). A repo-wide grep shows no test exercises this revert (`rg OperatorRpcAddressRequired test/` returns nothing). Add a `vm.expectRevert` test asserting the empty-rpcAddress call reverts with `OperatorRpcAddressRequired`, and a happy-path test confirming the event emits the supplied rpcAddress and storage stays empty.

🟡 **LOW** updateOperatorPreferences is not nonReentrant and fires a .call-based hook after state writes — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/core/Operators.sol)
> Unlike registerOperator/unregisterOperator, updateOperatorPreferences has no `nonReentrant` modifier. The BSM hook fires via `_tryCallManager` which uses a real `.call` (Base.sol:519), so a malicious manager can reenter during onUpdatePreferences. This PR actually IMPROVES the posture by moving the emit before the hook and ensuring all state writes (reg.updatedAt, prefs.ecdsaPublicKey, _blueprintOperatorKeys) complete before the hook — so a reentrant call observes fully-committed state. This is pre-existing (not introduced by the diff), flagged only because the reorder touches the interaction ordering. Consider adding `nonReentrant` for defense-in-depth parity with the register/unregister paths.

🟡 **LOW** New OperatorRpcAddressRequired error lacks natspec and is misplaced — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/libraries/Errors.sol)
> `OperatorRpcAddressRequired()` is an operator-related revert but is inserted between blueprint errors `BlueprintMetadataRequired()` and `BlueprintMembershipRequired()` under the BLUEPRINT section, and it has no natspec. This hurts discoverability and consistency. Fix: add a `/// @notice` docstring and move the error to the OPERATOR section.

🟡 **LOW** `OperatorRpcAddressRequired` is enforced on update but not on registration — asymmetric invariant — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/libraries/Errors.sol)
> The new error is declared here and used at Operators.sol:417 to make `updateOperatorPreferences` reject empty `rpcAddress` (rationale in-comment: an empty endpoint 'would otherwise broadcast an empty endpoint and break gossip reachability'). The same reachability argument applies to `_registerOperator` (Operators.sol:248-350), which still accepts empty `rpcAddress` and emits it in `OperatorRegistered` + forwards it to the BSM `onRegister` hook. So an operator can register with an empty endpoint (gossip-unreachable) and is only forced to supply one on the first update. This asymmetry is pre-existing (old code also permitted empty at registration), not a regression introduced by this PR, and the registration path is outside this shot's files — but the error declaration is what makes the inco

🟡 **LOW** Storage layout shift in `Blueprint` struct is safe only because protocol is pre-mainnet — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/libraries/Types.sol)
> Removing `uint32 operatorCount` from the middle of the packed slot-2 group (createdAt + operatorCount + membership + pricing + active = 15 bytes → createdAt + membership + pricing + active = 11 bytes) changes the byte offsets of `membership`, `pricing`, `active` within slot 2 of each `_blueprints[id]` value. For an in-place UUPS upgrade of a deployed contract this would misread those three fields (membership would read byte-0 of the old operatorCount, etc.). This is acceptable for THIS PR because deploy/LAUNCH-READINESS.md states the protocol is not yet launch-ready and has no mainnet deployment, so there is no live storage to migrate. The doc comment is accurate. Flagging only so the constraint is explicit: this struct change MUST NOT be shipped as an upgrade over any prior deployed Tangl

🟡 **LOW** `OperatorPreferences.rpcAddress` field retained but always empty — stale-reader footgun documented but not enforced — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/src/libraries/Types.sol)
> The field is kept in the struct (preserving storage layout of `_operatorPreferences`) but is now always written as `""` (Operators.sol:319) and the doc correctly directs consumers to read it from `OperatorRegistered`/`OperatorPreferencesUpdated` events instead. `getOperatorPreferences` (TangleCoreApi.sol:439-447) still returns the struct with an empty `rpcAddress`. Off-chain consumers (and the Rust bindings updated in this PR) that previously read `rpcAddress` from that getter will silently get empty strings. The doc at Types.sol:247-250 and TangleCoreApi.sol:434-438 covers this, so it is a documented breaking change, not a hidden bug. Nit: consider emitting the breaking-change note in bindings/CHANGELOG.md (not verified present) since the getter's effective return shape changed semantical

🟡 **LOW** Missing test coverage for new OperatorRpcAddressRequired revert — [`test/audit/medlow/Operators.t.sol`](https://github.com/tangle-network/tnt-core/blob/c88d239a6d123042e3fbcb146fdd58d7f8dad883/test/audit/medlow/Operators.t.sol)
> The changed test call sites now pass a non-empty rpcAddress to avoid the new revert in src/core/Operators.sol:417, but no test in this file (or the other changed test file test/Tangle.t.sol) asserts that updateOperatorPreferences(..., "", ...) reverts with OperatorRpcAddressRequired(). Add a dedicated test to cover this new requirement.

---
<sub>tangletools · 2026-07-02T13:35:24Z · [trace](https://gist.github.com/tangletools/b9bac04f6144877005769349b677baea)</sub>

tangletools (dismissed):
## ✅ Approved — 15 non-blocking findings — `c88d239a`

> Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/b9bac04f6144877005769349b677baea)

Summary comment for this run: [full summary](https://github.com/tangle-network/tnt-core/pull/194#issuecomment-4866324716)

---
<sub>tangletools · 2026-07-02T13:35:24Z · immutable [trace](https://gist.github.com/tangletools/b9bac04f6144877005769349b677baea)</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `7b6e42c7`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T19:16:24Z</sub>

tangletools (commented):
## 🟢 Value Audit — sound

| | |
|---|---|
| **Verdict** | sound |
| **Concerns** | 0 (none) |
| **Heuristic** | 0.1s |
| **Duplication** | 0.0s |
| **Interrogation** | 81.8s (2 bridge agents) |
| **Total** | 81.9s |

### 💰 Value — error

value agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

### 🎯 Usefulness — sound

Two clean storage cuts that event-source `rpcAddress` and derive operator count from the existing membership set; both are reachable and consumed exactly as the PR claims, and they remove a genuine drift-bug class.

- **Integration:** Both cuts land on already-reached surface. The canonical operator-count reader is `blueprintOperatorCount()` at src/core/TangleCoreApi.sol:485, now deriving from `_blueprintOperators[blueprintId].length()`; tests exercise it (test/Tangle.t.sol:48,166,219) and grep confirms NO on-chain reader touched the removed `Blueprint.operatorCount` struct field (rg `\.operatorCount` in src/ → zero hits; all `
- **Fit with existing patterns:** Fits the established grain. This is the identical discipline shipped in #193 (ba7b714 'hash+event-source the blueprint definition') applied to the next writers — same comment style, same event-only-with-BSM-hook-payload pattern, same 'denormalized duplicate of an EnumerableSet' removal rationale. The cached counter was the textbook drift-bug shape (two sources of truth for one fact); the set is al
- **Real-world viability:** Holds up better than the original on edge paths. The set-length read is O(1) and cannot drift; the old `bp.operatorCount++/--` could desync from the set on any future code path that touched one but not the other. The update path is hardened: `updateOperatorPreferences` now REQUIRES non-empty rpcAddress (src/core/Operators.sol:417, reverts `OperatorRpcAddressRequired`) precisely because storage is 
- **Bridge attempts:** 1

No concerns — sound change, no better or existing approach found. ✅

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T191949Z</sub>

tangletools:
## ❌ Needs Work — `7b6e42c7`

**Review health 100/100** · **Reviewer score 23/100** · **Confidence 95/100** · 18 findings (1 high, 2 medium, 15 low)

| | opencode-kimi | glm | aggregate |
|---|---|---|---|
| **Readiness** | 23 | 77 | **23** |
| **Correctness** | 23 | 77 | **23** |
| **Security** | 23 | 77 | **23** |
| **Testing** | 23 | 77 | **23** |
| **Architecture** | 23 | 77 | **23** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision.

### Blocking

🔴 **HIGH** UUPS storage layout corruption from Blueprint.operatorCount removal — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol)
> Types.sol (changed in this PR) removes `uint32 operatorCount` from the middle of the `Blueprint` struct, between `createdAt` and `membership`. Because Operators.sol, BlueprintsCreate.sol, and TangleCoreApi.sol read/write `_blueprints[blueprintId]` in storage, existing blueprints created before the upgrade will be misread: the 4 bytes that held `operatorCount` are now interpreted as the start of `membership`, `pricing`, and `active`. For typical small operator counts, existing blueprints will read as inactive with wrong membership/pricing, breaking service creation and pricing logic. The contract is UUPS upgradeable (Base.sol inherits UUPSUpgradeable), and no migration/reinitializer appears in the PR. Either keep `operatorCount` as a deprecated slot-filler, add an upgrade-time migration tha

### Other

🟠 **MEDIUM** TNT_CORE_VERSION points to a commit outside PR ancestry — [`bindings/TNT_CORE_VERSION`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/TNT_CORE_VERSION)
> The file now contains cd28a455a4b76982a6acfbc68659a1e711c25569, but `git merge-base --is-ancestor cd28a455a4b76982a6acfbc68659a1e711c25569 7b6e42c70804e96e87a77de839e8c172538252ac` returns 'not ancestor'. The reachable PR commit b84104318ebc8e104dd81bab9be867f10a50527c has the exact same tree (86305aefc576ec016f6fa4bb41a6e5b79e91f42f), so the bindings content is correct, but the recorded source hash cannot be checked out from the PR branch. This breaks reproducibility for downstream consumers who rely on TNT_CORE_VERSION. Fix: regenerate bindings so the file records a commit that is actually in the PR head's ancestry (e.g., b84104318ebc8e104dd81bab9be867f10a50527c, or 7b6e42c70804e96e87a77de839e8c172538252ac if regenerated after the final fix commit).

🟠 **MEDIUM** Blueprint struct storage/ABI break lacks local warning — [`src/libraries/Types.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/libraries/Types.sol)
> Removing `uint32 operatorCount` from `Blueprint` shifts `membership`, `pricing`, and `active` by 4 bytes within slot 2. Old layout: `createdAt(8) + operatorCount(4) + membership(1) + pricing(1) + active(1)`. New layout: `createdAt(8) + membership(1) + pricing(1) + active(1)`. Because `_blueprints` is a mapping stored in TangleStorage (UUPS-upgradeable architecture), an in-place upgrade would read stale `operatorCount` bytes as enum/bool values, corrupting every existing blueprint. TangleStorage.sol:14-22 already declares the layout greenfield/pre-mainnet and forbids in-place upgrades, but Types.sol gives no warning. Add an `@dev` note in Types.sol that this struct is NOT upgrade-compatible with earlier revisions and must only be used in fresh deployments.

🟡 **LOW** 3 ABI files carry stale-looking embedded source hash churn but no functional change — [`bindings/abi/IBlueprintServiceManager.json`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/abi/IBlueprintServiceManager.json)
> IBlueprintServiceManager.json, IMultiAssetDelegation.json, MultiAssetDelegation.json change ONLY the keccak256 of src/libraries/Types.sol embedded in compiler metadata (0x37cd036a… -> 0x47df918f…). Function/event/error sets are byte-identical (verified via set + ordered comparison). This is expected metadata noise from editing Types.sol and is harmless; noted only because a reviewer scanning the diff sees 1 +/- line per file with no visible semantic change.

🟡 **LOW** Unnecessary ABI regeneration churn — [`bindings/abi/IBlueprintServiceManager.json`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/abi/IBlueprintServiceManager.json)
> After normalizing ABI array order, this file is semantically identical to the base commit. It was regenerated alongside the other binding files but contributes only ordering/metadata noise to the diff. Avoid committing regenerated artifacts that did not actually change, or make the generator deterministic to prevent future spurious diffs.

🟡 **LOW** Unnecessary ABI regeneration churn — [`bindings/abi/IMultiAssetDelegation.json`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/abi/IMultiAssetDelegation.json)
> After normalizing ABI array order, this file is semantically identical to the base commit. The source changes do not touch IMultiAssetDelegation or MultiAssetDelegation, so this regeneration is pure diff churn.

🟡 **LOW** Breaking ABI change to Types.Blueprint tuple (getBlueprint) — off-chain consumers must re-encode — [`bindings/abi/ITangle.json`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/abi/ITangle.json)
> The getBlueprint return tuple loses its uint32 operatorCount component (field order: owner,manager,createdAt,[operatorCount removed],membership,pricing,active). Any off-chain/indexer/SDK consumer ABI-decoding getBlueprint with the OLD ABI will silently misalign every field after createdAt (membership, pricing, active) since tuple decoding is positional. This is the intentional subject of the PR (count is now derived via blueprintOperatorCount(uint64)->uint256), and the bindings here faithfully reflect the source. Flagging only so consumers (e.g. the Envio indexer in this repo, downstream Rust bindings) are confirmed regenerated against this new ABI before merge. No defect in the ABI file itself.

🟡 **LOW** Unnecessary ABI regeneration churn — [`bindings/abi/MultiAssetDelegation.json`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/bindings/abi/MultiAssetDelegation.json)
> After normalizing ABI array order, this file is semantically identical to the base commit. Regenerating unchanged implementation ABIs increases review surface without changing the interface.

🟡 **LOW** Breaking API change: updateOperatorPreferences with empty rpcAddress now reverts — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol)
> Previously, passing empty rpcAddress to updateOperatorPreferences meant 'keep unchanged' (the `if (bytes(rpcAddress).length > 0)` guard at the old line skipped the write). Now it reverts. Any external off-chain caller (operator bot, SDK, indexer tooling) that performs a key-only update with rpcAddress='' will break. The NatSpec documents this, but there is no migration/release note for external consumers. If external callers exist outside this repo, they need a coordinated update. All in-repo callers (tests) were updated to pass non-empty values.

🟡 **LOW** New OperatorRpcAddressRequired revert path has no test — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol)
> A new revert `if (bytes(rpcAddress).length == 0) revert Errors.OperatorRpcAddressRequired()` was added but no test exercises it. grep for OperatorRpcAddressRequired in test/ returns zero hits. Every test that calls updateOperatorPreferences passes a non-empty string. Add a vm.expectRevert test that calls updateOperatorPreferences with empty rpcAddress to lock in this behavior.

🟡 **LOW** New OperatorRpcAddressRequired revert path lacks explicit test coverage — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol)
> The changed tests (test/Tangle.t.sol and test/audit/medlow/Operators.t.sol) were updated to pass non-empty `rpcAddress` values, but no test asserts that `updateOperatorPreferences(blueprintId, key, "")` reverts with `Errors.OperatorRpcAddressRequired()`. Add a targeted regression test for this revert to prevent accidental regression.

🟡 **LOW** OperatorPreferencesUpdated event rpcAddress not asserted in tests — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol#L161)
> Since rpcAddress is no longer persisted to storage, the OperatorPreferencesUpdated event is the sole on-chain channel for the operator's endpoint. The existing test_UpdateOperatorPreferences (test/Tangle.t.sol:222) checks only updatedAt — it does not use vm.expectEmit to verify the event carries the correct rpcAddress. Given this is now the only way an indexer/operator learns the endpoint post-update, the event payload is the critical correctness invariant and should be explicitly asserted. By contrast, the registration path (test_RegisterOperator at [line 161](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol#L161)) does use vm.expectEmit for OperatorRegistered.

🟡 **LOW** registerOperator still permits empty rpcAddress while updateOperatorPreferences requires it — [`src/core/Operators.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/core/Operators.sol)
> `updateOperatorPreferences` now reverts if `rpcAddress` is empty because the value is event-only and an empty update would broadcast a missing endpoint. However, `_registerOperator` still allows an empty `rpcAddress` and emits `OperatorRegistered` with it. The same gossip-reachability rationale applies on first registration, so the API is inconsistent. Consider also requiring a non-empty `rpcAddress` in `_registerOperator`, or document why registration is intentionally allowed to omit it.

🟡 **LOW** New error missing NatSpec @notice, breaks file convention — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/libraries/Errors.sol#L67)
> Every other error in this file carries a `/// @notice` doc line (BlueprintMetadataRequired above it, BlueprintMembershipRequired below it). The newly added `error OperatorRpcAddressRequired();` at [line 67](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/libraries/Errors.sol#L67) has none, breaking the file's documentation invariant. Add: `/// @notice Operator update omitted a non-empty rpcAddress (required since rpcAddress is no longer persisted).` Pure doc/style; no behavioral impact.

🟡 **LOW** New operator error placed in blueprint section — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/libraries/Errors.sol)
> `OperatorRpcAddressRequired()` is inserted between `BlueprintMetadataRequired()` and `BlueprintMembershipRequired()` in the BLUEPRINT section, but it validates an operator-supplied RPC address in `Operators.updateOperatorPreferences`. This makes the error hard to locate and inconsistent with the file's organization. Move it to the OPERATOR section near `InvalidOperatorKey`/`DuplicateOperatorKey`.

🟡 **LOW** No test coverage for new OperatorRpcAddressRequired revert — [`src/libraries/Errors.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/src/libraries/Errors.sol)
> `Operators.updateOperatorPreferences` now reverts with `OperatorRpcAddressRequired()` when `rpcAddress` is empty (Operators.sol:417). Grep across `test/` finds no test referencing `OperatorRpcAddressRequired`, and the only test updates in this PR changed existing calls from `rpcAddress=""` to a non-empty URL to avoid the revert. Add a test that calls `updateOperatorPreferences(blueprintId, key, "")` and asserts the revert to prevent regression.

🟡 **LOW** Missing coverage for OperatorRpcAddressRequired revert — [`test/Tangle.t.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/test/Tangle.t.sol#L228)
> The PR's source change in src/core/Operators.sol now requires a non-empty rpcAddress in updateOperatorPreferences (reverting with Errors.OperatorRpcAddressRequired). test_UpdateOperatorPreferences at [line 228](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/test/Tangle.t.sol#L228) happens to pass 'newPrefs', so it does not break, but the file adds no test that an empty rpcAddress reverts. Add a test like test_UpdateOperatorPreferences_RevertEmptyRpcAddress that calls updateOperatorPreferences(blueprintId, '', '') and expect-reverts OperatorRpcAddressRequired.

🟡 **LOW** Missing explicit test for new OperatorRpcAddressRequired revert — [`test/audit/medlow/Operators.t.sol`](https://github.com/tangle-network/tnt-core/blob/7b6e42c70804e96e87a77de839e8c172538252ac/test/audit/medlow/Operators.t.sol)
> The PR introduces a new revert OperatorRpcAddressRequired in Operators.sol:417 when updateOperatorPreferences receives an empty rpcAddress. The shot updates existing key-swap tests to pass a non-empty URL so they do not hit this revert, but it does not add a dedicated test asserting that empty rpcAddress reverts with the expected selector. Adding such a test would cover the new revert condition and prevent future regressions.

---
<sub>tangletools · 2026-07-02T19:51:14Z · [trace](https://gist.github.com/tangletools/67f2c539e6ff9c4d2404fb28e6ce51e5)</sub>

tangletools (changes requested):
## ❌ 1 Blocking Finding — `7b6e42c7`

> Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 7/7 planned shots over 16 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/67f2c539e6ff9c4d2404fb28e6ce51e5)

Summary comment for this run: [full summary](https://github.com/tangle-network/tnt-core/pull/194#issuecomment-4869911190)

---
<sub>tangletools · 2026-07-02T19:51:14Z · immutable [trace](https://gist.github.com/tangletools/67f2c539e6ff9c4d2404fb28e6ce51e5)</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `02911113`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-02T19:57:56Z</sub>

tangletools (commented):

| | |
|---|---|
| **Concerns** | 1 (1 weak-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 154.1s (2 bridge agents) |
| **Total** | 154.1s |

### 💰 Value — sound-with-nits

Event-sources operator rpcAddress and derives blueprint operatorCount from the EnumerableSet, cutting redundant SSTOREs while keeping on-chain invariants intact; a good, coherent follow-through of the #193 discipline, with one stale interface doc to fix.

- **What it does:** Two storage-to-event moves in the operator lifecycle. (1) `rpcAddress` is no longer persisted in `_operatorPreferences`; it is carried by `OperatorRegistered` / `OperatorPreferencesUpdated` and the BSM hook payload (`getOperatorPreferences().rpcAddress` now returns `""`). `updateOperatorPreferences` now requires a non-empty `rpcAddress` (new `OperatorRpcAddressRequired` error) because there is no 
- **Goals it achieves:** Reduce lifecycle gas on high-SSTORE-metering chains (notably Tempo, where the unbounded-string SSTORE was ~120–220k gas per register/update). Remove denormalized state that can drift from its canonical source. Extend the store-hash/emit-event deduplication discipline from #193 to the next-heaviest lifecycle writers without touching the economic core.
- **Assessment:** Good change on its merits. The removed data had no on-chain readers (`rpcAddress` is consumed off-chain for gossip discovery; `operatorCount` duplicated the existing `_blueprintOperators` set). Deriving the count from `EnumerableSet.length()` is idiomatic and removes a bug class. The review follow-up (require non-empty `rpcAddress`, CEI emit order) correctly closes the edge case where a key-only u
- **Better / existing approach:** none — this is the right approach. The codebase already established the event-source/digest pattern in #193 (`ba7b714`, `src/core/BlueprintsCreate.sol` and `src/libraries/Types.sol`); this PR applies it consistently to operators. A more aggressive refactor could drop `rpcAddress` from `Types.OperatorPreferences` entirely, but that would break the BSM hook payload ABI and the `getOperatorPreference
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

### 🎯 Usefulness — sound

Clean gas optimization that event-sources rpcAddress and derives the blueprint operator count from the canonical membership set, with no broken consumers and a real drift-bug class eliminated.

- **Integration:** Fully wired and reachable. The dropped Blueprint.operatorCount field had ZERO in-repo production readers — grep across all .sol confirms every remaining `.operatorCount++/--` is on `svc` (Service) or `stats` (JobsAggregation local), not `bp`; the one test read (test/Tangle.t.sol:48) was migrated to `blueprintOperatorCount()`, which already derives from `_blueprintOperators[id].length()` (TangleCor
- **Fit with existing patterns:** Follows the exact store-hash/emit-event + derive-from-canonical-set discipline that #193 established for blueprint definitions — this PR applies it to the next lifecycle writers. Consistent with the codebase's grain; competes with no existing pattern. The follow-up `OperatorRpcAddressRequired` guard (Operators.sol:417) and CEI emit-before-hook ordering (Operators.sol:453-454) match the contract's 
- **Real-world viability:** Holds up off the happy path. The empty-rpcAddress guard closes the genuine footgun an event-sourced endpoint creates (a key-only update would otherwise broadcast an empty endpoint and break gossip dial-out). Deriving the count from the set removes the drift class where the cached counter and `_blueprintOperators` could disagree. The first-join metadata-lock re-keyed from `++count == 1` to `.length
- **Bridge attempts:** 1

### 💰 Value Audit

🟡 **Stale interface NatSpec contradicts new rpcAddress requirement** `[maintenance]` ``
> `src/interfaces/ITangleOperators.sol:31` says `rpcAddress` in `OperatorPreferencesUpdated` 'may be empty if unchanged', and `:67` tells callers to 'pass empty string to keep unchanged'. Both are now wrong: `Operators.updateOperatorPreferences` reverts with `OperatorRpcAddressRequired()` when `rpcAddress` is empty (`src/core/Operators.sol:417`). The interface docs should be updated to match the implementation so integrators and the Rust SDK don't build against the old semantics.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260702T201226Z</sub>

PR #196, opened 2026-07-03
feat(indexer): decode 0.18 event-sourced blueprint definitions + binary versions

drewstone:
## Problem
On 0.18 chains `getBlueprintDefinition` storage is zeroed — `BlueprintDefinitionRecorded` is the ONLY on-chain source of blueprint/job display prose — but the indexer stored the payload as raw hex (names/descriptions unqueryable), never indexed `BinaryVersionRecorded` at all, and the redeploy sync script could not set the MBSM address (manifests don't carry it), so a standard 0.18 cutover would silently leave blueprint display data pointed at a dead contract.

## Solution
- **Decode the definition**: viem decode with an ABI parameter spec extracted from `bindings/abi/ITangleFull.json` (machine-generated). Typed `{succeeded, value, error}` outcome; decode failure logs loudly + lands in a `decodeError` field; raw hex always persisted. Display fields materialized on `BlueprintDefinition` and mirrored (name/description) onto `Blueprint` for join-free catalog queries.
- **BinaryVersionRecorded** → new `BinaryVersion` entity (both configs + handler), signature verified against the emit site (`MasterBlueprintServiceManager.sol:76`).
- **sync-config-from-manifest**: requires `--mbsm 0x…` (or an `mbsm` manifest key) and updates the MBSM block — hard error instead of the silent-dead-address trap.

## Verification
```
pnpm test        # 4 files, 7 passed — incl. a cross-implementation golden vector:
                 # the fixture is abi.encode(def) produced by FORGE (Solidity encoder),
                 # decoded by viem — all 9 display fields + 8 jobs exact
pnpm typecheck   # green (after envio codegen)
sync script      # fail-loud proven (tempo manifest, no mbsm → hard error);
                 # success path proven (base-sepolia manifest + --mbsm writes the block)
```

## Notes
- Re-index from a FRESH database at 0.18 cutover — blueprint/service IDs reset on greenfield redeploys; mixing entities across deployments corrupts ID-keyed state.
- Emitter trust = the configured MBSM address (envio address filter), same anchor as every other contract here; `keccak(encodedDefinition) == blueprintDefinitionHash(id)` remains the integrity check for untrusted readers.


tangletools (dismissed):
## ✅ Auto-approved drewstone PR — `a7e5cfc8`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-03T03:47:23Z</sub>

tangletools (commented):
## 🟠 Value Audit — better-approach-exists

| | |
|---|---|
| **Verdict** | better-approach-exists |
| **Concerns** | 1 (1 medium-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 136.9s (2 bridge agents) |
| **Total** | 136.9s |

### 💰 Value — better-approach-exists

Decodes the event-sourced blueprint definition blob + indexes BinaryVersionRecorded + fail-closes the MBSM sync gap — coherent and correct, but hand-transcribes a 257-line ABI spec the repo already ships as machine-generated JSON.

- **What it does:** Completes the indexer side of the 0.18 protocol change (#193) that moved blueprint display data out of storage into events. Three deltas: (1) decodes the raw `abi.encode(Types.BlueprintDefinition)` blob carried by `BlueprintDefinitionRecorded` and materializes 9 display fields + job name/description arrays onto `BlueprintDefinition`, mirroring name/description onto `Blueprint` for join-free catalo
- **Goals it achieves:** Make 0.18 blueprint display data (names, descriptions, job prose) queryable in the indexer at all — without this it's raw hex blobs, since #193 zeroed the storage fields. Index binary-version lifecycle events that had no indexer representation. Make the post-redeploy cutover fail loudly instead of silently breaking catalog queries. All three serve 0.18 cutover readiness for the indexer.
- **Assessment:** Sound and in-grain on the whole. Event signatures, field names, and the decode struct were all verified against src/ — `BinaryVersionRecorded`'s params (blueprintId/versionId/sha256Hash/binaryUri) match the handler reads exactly; the struct in blueprintDefinitionAbi.ts matches Types.BlueprintDefinition. The typed `{succeeded, value, error}` decode outcome, raw-hex-always-persisted invariant, loud 
- **Better / existing approach:** Searched for existing decode logic: only an unrelated service-request decoder in tangle.ts (no duplication). Searched for whether the repo already has the struct as a machine-readable ABI: yes — bindings/abi/ITangleFull.json (referenced in the new file's own header comment as the extraction source). The struct could be loaded from there as the single source of truth instead of hand-transcribed.
- **Bridge attempts:** 2
- **Bridge warning:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content

### 🎯 Usefulness — sound

Decodes the only on-chain source of blueprint display prose on 0.18 chains and indexes BinaryVersionRecorded for the first time, wired exactly in the codebase's grain with fail-loud error paths.

- **Integration:** Fully reachable: both new handlers are registered against the deployed MasterBlueprintServiceManager in config.yaml (events match the Solidity emit sites at MasterBlueprintServiceManager.sol:64,76 verbatim); the BlueprintDefinition/BinaryVersion entities are consumed via the standard context.Entity.set pattern; the Blueprint mirror-write keys off blueprintId.toString() matching BlueprintCreated in
- **Fit with existing patterns:** Follows the established envio handler pattern precisely — same onEvent shape, same handlerUtils (getEventId/toBigInt/toHexString/normalizeAddress), and the same read-modify-write Blueprint.get→spread→set flow used by BlueprintUpdated/BlueprintTransferred in tangle.ts. No competing decoder exists; this completes the raw-hex-stored-by-PR-#193 path rather than forking anything.
- **Real-world viability:** Holds up off the happy path: decode failure is fail-loud (error logged + decodeError field populated + raw hex still persisted), the result type is a discriminated union so callers can't silently ignore failure, BinaryVersion is keyed `${blueprintId}-${versionId}` for idempotent upsert matching the on-chain append-only semantics, and the sync script hard-errors instead of silently keeping a stale 
- **Bridge attempts:** 1

### 💰 Value Audit

🟠 **257-line ABI spec hand-transcribed from a JSON the repo already ships** `[maintenance]` ``
> blueprintDefinitionAbi.ts is a hand-maintained copy of Types.BlueprintDefinition, transcribed from bindings/abi/ITangleFull.json (the file's own header at lines 5-8 says so). The entire decodeError field + loud error log + fixture-regen-by-hand instructions exist primarily to absorb the drift risk this transcription creates: when Types.BlueprintDefinition changes on-chain, a developer must remember to re-transcribe the spec AND regenerate the fixture, or decode silently breaks at runtime. The ma

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260703T041847Z</sub>

drewstone:
Addressed the value-audit concern (hand-transcribed ABI spec drift risk): the spec must stay checked in because envio's bundler can't import JSON from outside the indexer package — but drift is now CI-enforced instead of remembered. `blueprintDefinitionAbi.test.ts` re-derives the spec from `bindings/abi/ITangleFull.json` at test time and deep-compares against the checked-in constant; a struct change reds the suite with instructions to regenerate both the spec and the Solidity fixture. Mutation-verified (renaming one field in the spec fails the test).

tangletools (approved):
## ✅ Auto-approved drewstone PR — `363df203`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-03T04:23:50Z</sub>

PR #204, opened 2026-07-07
perf(0.19): hot-path gas wins + security hardening from deep pre-testnet audit

drewstone:
Final efficiency + security pass before the testnet cut, folded into 0.19 (bindings unpublished, so no extra downstream churn). A per-subsystem deep audit (gas + security, shrink-regression + fresh adversarial, every finding independently verified) found **22 hot-path gas wins** and **9 security items — all low/info**. The 0.19 shrink introduced **no real bug** (both "shrink-regression" findings were info: a deploy guardrail + a dead event decl).

## Gas (runtime, hot paths)
- **Storage packing** (`Types.sol`): reorder `BondInfoDelegator` 4→3, `BondLessRequest` 5→4, `JobCall` 4→3, `BinaryVersion` 4→3, `AssetConfig` 6→5 slots → **one cold SSTORE (~20k gas) saved per** new delegation / scheduleUnstake / job submit / version publish / asset write. Field types/names unchanged; named construction is order-independent so no caller changes.
- **Billing**: cache `ServiceOperator.active` (1 SLOAD/op not 2); thread the `(op,asset)` keccak forward so cursor-commit reuses it; compute the bond asset once per bill not 3×.
- **Staking**: reuse the just-written `delegatedAmount`; cache array ref+length; `unchecked` bounded counters.
- **Jobs**: hoist per-service reads out of per-quote loops; single operator-set snapshot per aggregated submit; `unchecked` counters.
- **Rewards**: cache `StreamingPayment` fields in the per-claim `_drip`; hoist repeated `_totalFixedScoreByAsset` SLOADs.

## Security
- **USD-scale fail-closed**: a subscription leg that's USD-pinned but has no price snapshot and no oracle now contributes **0** instead of silently billing at 1× (was a slow under-bill). Valid-oracle behavior unchanged.
- **`disputeSlash`**: hoist the cheap status/window guards above the blueprint-manager staticcalls (griefing-cost defense).
- Remove the dead `BlueprintSourcesUpdated` decl left in `ITangleBlueprints` (0.19 shrink cleanup).

## Discipline note
During implementation the agents **rejected two of the audit's own suggestions** as unsafe (an escrow pre-filter would break the happy path — the QoS hook can rescue a bill to ≤ escrow; a "USD pre-pass dedup" wasn't a real duplicate). Skipped, not forced.

## Verification
- **810 money-path tests pass** (payments/billing/subscription, slashing, delegation/staking, jobs/RFQ/aggregation, rewards/streaming, E2E/full-stack/integration).
- src compiles (one stack-too-deep from over-caching fixed by reading `req.shares` inline — kept every other win).
- Bindings regenerated for the reordered struct tuples + removed dead event.

## Deployment guardrail (from the audit, INFO)
Before promoting the combined 0.19 build to a proxy: confirm no live proxy was initialized on the **pre-shrink storage layout** (the mid-layout removals + struct reorders are only layout-safe because the target is undeployed). For a fresh testnet deploy this is a non-issue.


tangletools (approved):
## ✅ Auto-approved drewstone PR — `46ff9274`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-07T20:17:38Z</sub>

PR #212, opened 2026-07-24
chore(bindings): republish 0.19.1 with the deployed 0.19 ABI (0.19.0 was stale)

drewstone:
The crates.io `tnt-core-bindings 0.19.0` is **stale vs the deployed 0.19 contract** and needs a corrective republish before operators can install from published artifacts instead of git patches.

## Why
- `0.19.0` was published **2026-07-07 22:24**.
- **#209** (EventDriven per-job billing in the service settlement asset) merged **2026-07-09 04:39** and regenerated the ABI (`i_tangle.rs` +879, `i_tangle_full.rs` +921, …).
- So published `0.19.0` ships an **intermediate 0.19 ABI that was never deployed** — its `i_tangle.rs` differs from main by ~900 lines (diffed the extracted crate to confirm).
- Consumers pinning `0.19.0` — **including blueprint-sdk** (`tnt-core-bindings = "0.19"`) — build against selectors that don't match the live 0.19 deployment. The operator worked on Tempo only because it consumed bindings via a git-pin override, which masked the staleness.

## Change
- `bindings/Cargo.toml`: `0.19.0` → `0.19.1` (+ `Cargo.lock`). Carries main's current, deployed-matching ABI, and folds in the MSRV-declaration fix from #211.
- CHANGELOG documents what `0.19.0` was missing (the #209 payments ABI).

## Verification
`cargo publish --dry-run -p tnt-core-bindings --locked` → packages 36 files, compiles clean in isolation on Rust 1.91; upload aborted only by `--dry-run`.

## After merge
Publish `0.19.1` to crates.io, then bump blueprint-sdk's `tnt-core-bindings` dep to `0.19.1` and release it (unblocks the queued blueprint `chore: release` #1470) so operators install the aligned binary.

PR #211, opened 2026-07-24
fix(bindings): declare the real MSRV (rust-version 1.81 → 1.91)

drewstone:
Follow-up to #198, closing its multi-shot review's sole blocking finding ([review comment](https://github.com/tangle-network/tnt-core/pull/198#issuecomment-4883923685)).

## The bug
#198's security sweep bumped `alloy` 1.1.2 → 1.8.3 — **load-bearing**: the `lru` 0.16 RUSTSEC fix requires `alloy-provider` ≥1.4, which moves its `lru` pin `^0.13` → `^0.16`, which requires `alloy` 1.8.3. And `alloy` 1.8.3 declares `rustc >= 1.91`. But `bindings/Cargo.toml` and `fixtures/Cargo.toml` still declared `rust-version = "1.81"`.

`tnt-core-bindings` is a **published** crate. Advertising `1.81` while the locked graph needs `1.91` means a consumer on Rust 1.81–1.90 hits confusing `rustc X.Y is unsupported` errors from alloy/ruint instead of a clear MSRV error.

I verified the true max MSRV is **1.91** (`cargo metadata` over the locked graph — every `alloy` 1.8.3 sub-crate declares 1.91).

## The fix
- `rust-version` `1.81` → `1.91` in `bindings/Cargo.toml` + `fixtures/Cargo.toml` — the declaration now matches what the graph requires.
- `bindings/CHANGELOG.md` documents the MSRV bump **and** the two build-environment consequences of the required alloy bump, as a conscious record (the reviewer asked these be a deliberate decision, not an accidental `cargo update` side effect):
  - **cmake + a C toolchain** are now build-time requirements (`aws-lc-rs`/`aws-lc-sys` via `reqwest` 0.13 → `rustls` 0.23.41).
  - **TLS trust root** moved from bundled `webpki-roots` to the OS-native store (`rustls-platform-verifier`).

## Why not cap alloy / force `ring`
- **Capping alloy is off the table**: it would reintroduce the `lru` advisory that #198 fixed.
- **Not forcing the `ring` crypto backend** (to drop cmake): alloy 1.8.3 defaults to `aws-lc-rs`; overriding a transitive crypto provider from a downstream crate is fragile (additive Cargo features, controlled by alloy/reqwest), CI already has cmake, and the protocol isn't live. Documenting the requirement is the right call; dropping cmake later is a deliberate alloy-feature change, not a security follow-up.

[...]

PR #209, opened 2026-07-09
feat(payments): EventDriven per-job billing in the service's settlement asset (native or ERC20)

drewstone:
## Summary
Generalizes EventDriven per-job billing to settle in the service's **developer-declared settlement asset** — native OR a manager-allowed ERC20 (e.g. Tempo's PathUSD) — instead of a native-`msg.value`-only hardcode. This unblocks paid jobs on chains that forbid native value transfers (Tempo bans them at the VM level), while keeping native byte-identical on Tangle L1s.

## Why
`PaymentLib.collectPayment` already handles both native and hardened ERC20 (fee-on-transfer rejection, `msgValue==0` enforced); `requestService`/escrow already carry a `paymentToken`. Only the EventDriven per-job path hardcoded `address(0)`, and one rule forbade a token — so pay-per-job services could not exist on a value-banning chain. This is a **consistency fix reusing audited primitives**, not new payment machinery.

## Design (the key safety decision)
The settlement asset is chosen by the **developer at the blueprint level** (`_blueprintSettlementAsset[blueprintId]`, owner-gated `setBlueprintSettlementAsset`, allow-list-validated), and **pinned to the service at activation** — never chosen by the customer. This binds the asset and the per-job rate to the same party in the same units, so a decimal/asset mismatch is structurally impossible. Rate is denominated in the settlement asset's smallest unit (documented on the setter/view).

## Security process — two adversarial-review rounds
- **Round 1 found a real HIGH:** v1 pinned the asset from the *customer's* request while the rate is per-blueprint → a customer could request native against a 6-dec PathUSD rate and settle every job at ~1e-12× the intended price (10^12× underpayment), with the base BSM unable to deny native.
- **Fix:** moved asset choice to the developer/blueprint, pinned at activation, re-closed `requestService` for EventDriven (customer token must be `address(0)`), and closed a **second live path** — quote-created services (`createServiceFromQuotes`) also never pinned the asset.
- **Round 2: `merge_safe: true`, zero confirmed issues** across collect-vs-distribute asset consistency, storage-append safety, fail-closed asset authorization (4 layers), reentrancy/CEI (global OZ guard holds across the new ERC20 callbacks; pull-pattern payouts), and native back-compat.

## Verification
[...]

PR #213, opened 2026-07-24
ci(bindings): fail a PR that changes the ABI without bumping the bindings version

drewstone:
Closes the drift class that shipped a stale published crate and silently broke operators.

## What went wrong (this is not hypothetical)
- `tnt-core-bindings 0.19.0` was published to crates.io **2026-07-07**.
- #209 regenerated the ABI **2026-07-09** — under the **same** version number.
- crates.io versions are immutable, so the published `0.19.0` kept an ABI **~900 lines** out of date from the deployed contract.
- `blueprint-sdk` (and every operator built from it) linked selectors that don't exist on-chain. **Nothing failed loudly** — the live Tempo operator only worked because it consumed bindings through a `[patch.crates-io]` git pin that masked the staleness.
- Fixed after the fact by republishing as `0.19.1` (#212). tnt-core CI had **no** check that would have caught it.

## The gate
Pure git, no build, runs in seconds: if a PR touches `bindings/abi/**` or `bindings/src/bindings/**`, then the `version` literal in `bindings/Cargo.toml` must differ from the base commit's. It compares the *literal* on both sides, so an unrelated edit to that file can't satisfy the gate.

## Validated against real history
| PR | ABI changed? | version bumped? | gate |
|---|---|---|---|
| #209 (payments) | yes | **no** | ❌ **fires** — the actual bug |
| #211 (MSRV only) | no | n/a | N/A — correctly silent |

## Scope
One new workflow file. Deliberately does **not** regenerate bindings in CI to diff them: `gen-bindings` runs a full `forge build`, and per `foundry.yml`'s header the cold via-IR compile OOM-kills a stock runner. This guard is the cheap, precise half; a regeneration-diff job needs a larger runner and is a separate follow-up.

### DIFF
--- a/.github/workflows/bindings-guard.yml
+++ b/.github/workflows/bindings-guard.yml
@@ -1,0 +1,72 @@
+# Bindings publish guard.
+#
+# The failure this exists to prevent (it happened, and it silently broke
+# operators): `bindings/` was regenerated when the contracts changed, but
+# `bindings/Cargo.toml`'s version stayed the same. crates.io already had that
+# version, so the published crate kept an ABI that no longer matched the
+# deployed contract — `tnt-core-bindings 0.19.0` was published 2026-07-07 and
+# #209 changed the ABI on 2026-07-09 under the same version number, leaving the
+# published crate ~900 lines out of date. Consumers (blueprint-sdk, and every
+# operator built from it) linked selectors that do not exist on-chain, and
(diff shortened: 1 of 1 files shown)


tangletools (approved):
## ✅ Auto-approved drewstone PR — `c020ffb8`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-24T23:49:15Z</sub>

PR #217, opened 2026-08-14
fix(aggregation): honor EIP-197 G2 coordinate order

drewstone:
Risk class: D — cryptographic correctness fix.

## Problem

Multi-signer BLS aggregation reverted with `AggregatedPubkeyMismatch()`.

The library stored G2 coordinates in EIP-197 order `[imaginary, real]`, but its Fp2 multiplication and inversion treated index zero as the real component.

The test helper also had permuted `2*G2`, `3*G2`, and placeholder `6*G2` vectors.

## Fix

- Make Fp2 multiplication and inversion operate on the stored `[imaginary, real]` order.
- Replace helper vectors with values generated by Arkworks `ark-bn254 0.4.0`.
- Add add/double regression vectors for `2*G2`, `3*G2`, and `6*G2`.
- Replace the skipped three-signer test with real two- and three-signer on-chain flows.

## Proof

The old implementation fails the new two-key vector:

```text
FOUNDRY_PROFILE=fast forge test --match-contract BN254G2ArithmeticTest --match-test test_add_two_operator_keys_matches_arkworks_twoSignerAggregate -vv
[FAIL: 2G2 x[0]: 2801474874848198952174774680213679937838371328029144741324069556084792504510 != 14583779054894525174450323658765874724019480979794335525732096752006891875705]
```

The fixed implementation passes all four Arkworks arithmetic vectors.

- `BN254G2ArithmeticTest`: 4 passed.
- `BLSAggregationE2ETest`: 12 passed, including two signers at 1,822,200 gas and three signers at 4,904,253 gas.
- `BLSAggregation.t.sol`: 36 passed.
- `BN254Infinity.t.sol`: 2 passed.
- `FOUNDRY_PROFILE=local_build forge build`: passed.

`FOUNDRY_PROFILE=local_build forge build --sizes` still reports existing over-limit non-deployment contracts, including `GovernanceDeployer` at 86,517 bytes. This change does not affect that baseline.

## Compatibility

No public ABI, selector, or storage layout changes. The bindings remain at version `0.19.1`.


tangletools (approved):
## ✅ Auto-approved drewstone PR — `fe62cf12`

**This approval is provisional.** It rests on the audit running. If the audit cannot run — for example the CLI bridge rejects it — this approval is dismissed rather than left standing, so an unrun check never reads as a passing one.

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-08-14T10:20:42Z</sub>

